PR #25, opened 2024-04-17
feat(api): add array

minghuaw:
This is an initial implementation of the safe rust wrapper of mlx array. 

Below are some problems found that should probably be discussed (and hopefully resolved)

1. The c api doesn't expose c++ exceptions, which would cause a fatal runtime error 
    ```sh
    fatal runtime error: Rust cannot catch foreign exceptions
    ```
2. The `Array` type should probably NOT implement the `Clone` trait? Because the c struct mlx_array is atomically reference counted but not protected by `Mutex`
3. A good type alias for `num_complex::Complex<f32>`, `c64` (the current one) or `complex64`. `Complex64` should probably be avoided because `num_complex::Complex64` is actually `num_complex::Complex<f64>`

There is one thing that we probably want to double check after implementing ops

For `mlx_array_data_*` methods, the mlx doc says

> Array must be evaluated, otherwise returns NULL.

### DIFF
--- a/src/array.rs
+++ b/src/array.rs
@@ -200,6 +200,12 @@ impl Drop for Array {
 }
 
 impl Array {
+    /// Create a new array from an existing mlx_array pointer.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure the reference count of the array is properly incremented with
+    /// `mlx_sys::mlx_retain`.
     pub unsafe fn from_ptr(ptr: mlx_array) -> Array {
         Array { ptr }
     }
--- a/src/array.rs
+++ b/src/array.rs
@@ -12,7 +12,7 @@ use crate::sealed::Sealed;
 pub type complex64 = Complex<f32>;
 
 /// Array element type
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[repr(u32)]
 pub enum Dtype {
     Bool = mlx_sys::mlx_array_dtype__MLX_BOOL,
@@ -246,7 +246,7 @@ impl Array {
     /// - `data`: A buffer which will be copied.
     /// - `shape`: Shape of the array.
     /// - `dim`: Number of dimensions (size of shape).
-    pub fn from_data<T: ArrayElement>(data: &[T], shape: &[i32], dim: i32) -> Self {
+    pub fn from_slice<T: ArrayElement>(data: &[T], shape: &[i32], dim: i32) -> Self {
         let c_array = unsafe {
             mlx_sys::mlx_array_from_data(
                 data.as_ptr() as *const c_void,
@@ -331,7 +331,7 @@ impl Array {
     }
 
     /// Returns a pointer to the array data
-    pub fn data<T: ArrayElement>(&self) -> &[T] {
+    pub fn as_slice<T: ArrayElement>(&self) -> &[T] {
         // TODO: check below after ops are implemented
         // Array must be evaluated, otherwise returns NULL.
 
@@ -419,8 +419,8 @@ mod tests {
     #[test]
     fn new_array_from_single_element_slice() {
         let data = [1i32];
-        let array = Array::from_data(&data, &[1], 1);
-        assert_eq!(array.data::<i32>(), &data);
+        let array = Array::from_slice(&data, &[1], 1);
+        assert_eq!(array.as_slice::<i32>(), &data);
         assert_eq!(array.item::<i32>(), 1);
         assert_eq!(array.item_size(), 4);
         assert_eq!(array.size(), 1);
@@ -435,8 +435,8 @@ mod tests {
     #[test]
     fn new_array_from_multi_element_slice() {
         let data = [1i32, 2, 3, 4, 5];
-        let array = Array::from_data(&data, &[5], 1);
-        assert_eq!(array.data::<i32>(), &data);
+        let array = Array::from_slice(&data, &[5], 1);
+        assert_eq!(array.as_slice::<i32>(), &data);
         assert_eq!(array.item_size(), 4);
         assert_eq!(array.size(), 5);
         assert_eq!(array.strides(), &[1]);
@@ -452,7 +452,7 @@ mod tests {
     // #[should_panic]
     // fn get_item_from_multi_element_array_should_panic() {
     //     let data = [1, 2, 3, 4, 5];
-    //     let array = Array::from_data(&data, &[5], 1);
+    //     let array = Array::from_slice(&data, &[5], 1);
     //     array.item::<i32>();
     // }
 }
--- a/src/array.rs
+++ b/src/array.rs
@@ -9,10 +9,10 @@ use crate::sealed::Sealed;
 // TODO: camel case?
 // Not using Complex64 because `num_complex::Complex64` is actually Complex<f64>
 #[allow(non_camel_case_types)]
-pub type c64 = Complex<f32>;
+pub type complex64 = Complex<f32>;
 
 /// Array element type
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, PartialEq, Eq)]
 #[repr(u32)]
 pub enum Dtype {
     Bool = mlx_sys::mlx_array_dtype__MLX_BOOL,
@@ -155,22 +155,22 @@ impl ArrayElement for bf16 {
     }
 }
 
-impl Sealed for c64 {}
+impl Sealed for complex64 {}
 
-impl ArrayElement for c64 {
+impl ArrayElement for complex64 {
     const DTYPE: Dtype = Dtype::Complex64;
 
     fn scalar_array_item(array: &Array) -> Self {
-        let bindgen_c64 = unsafe { mlx_sys::mlx_array_item_complex64(array.ptr) };
+        let bindgen_complex64 = unsafe { mlx_sys::mlx_array_item_complex64(array.ptr) };
 
         Self {
-            re: bindgen_c64.re,
-            im: bindgen_c64.im,
+            re: bindgen_complex64.re,
+            im: bindgen_complex64.im,
         }
     }
 
     fn array_data(array: &Array) -> *const Self {
-        // c64 has the same memory layout as __BindgenComplex<f32>
+        // complex64 has the same memory layout as __BindgenComplex<f32>
         unsafe { mlx_sys::mlx_array_data_complex64(array.ptr) as *const Self }
     }
 }
@@ -234,7 +234,7 @@ impl Array {
     }
 
     /// New array from a complex scalar.
-    pub fn from_complex(val: c64) -> Array {
+    pub fn from_complex(val: complex64) -> Array {
         let ptr = unsafe { mlx_sys::mlx_array_from_complex(val.re, val.im) };
         Array { ptr }
     }
@@ -260,7 +260,7 @@ impl Array {
     }
 
     /// The size of the array’s datatype in bytes.
-    pub fn itemsize(&self) -> usize {
+    pub fn item_size(&self) -> usize {
         unsafe { mlx_sys::mlx_array_itemsize(self.ptr) }
     }
 
@@ -317,6 +317,7 @@ impl Array {
         Dtype::try_from(dtype).unwrap()
     }
 
+    // TODO: document that mlx is lazy
     /// Evaluate the array.
     pub fn eval(&mut self) {
         // This clearly modifies the array, so it should be mutable
@@ -325,6 +326,7 @@ impl Array {
 
     /// Access the value of a scalar array.
     pub fn item<T: ArrayElement>(&self) -> T {
+        // TODO: check and perform type conversion from the inner type to the desired output type
         T::scalar_array_item(self)
     }
 
@@ -365,7 +367,7 @@ mod tests {
     fn new_scalar_array_from_bool() {
         let array = Array::from_bool(true);
         assert_eq!(array.item::<bool>(), true);
-        assert_eq!(array.itemsize(), 1);
+        assert_eq!(array.item_size(), 1);
         assert_eq!(array.size(), 1);
         assert!(array.strides().is_empty());
         assert_eq!(array.nbytes(), 1);
@@ -378,7 +380,7 @@ mod tests {
     fn new_scalar_array_from_int() {
         let array = Array::from_int(42);
         assert_eq!(array.item::<i32>(), 42);
-        assert_eq!(array.itemsize(), 4);
+        assert_eq!(array.item_size(), 4);
         assert_eq!(array.size(), 1);
         assert!(array.strides().is_empty());
         assert_eq!(array.nbytes(), 4);
@@ -391,7 +393,7 @@ mod tests {
     fn new_scalar_array_from_float() {
         let array = Array::from_float(3.14);
         assert_eq!(array.item::<f32>(), 3.14);
-        assert_eq!(array.itemsize(), 4);
+        assert_eq!(array.item_size(), 4);
         assert_eq!(array.size(), 1);
         assert!(array.strides().is_empty());
         assert_eq!(array.nbytes(), 4);
@@ -402,10 +404,10 @@ mod tests {
 
     #[test]
     fn new_scalar_array_from_complex() {
-        let val = c64 { re: 1.0, im: 2.0 };
+        let val = complex64 { re: 1.0, im: 2.0 };
         let array = Array::from_complex(val);
-        assert_eq!(array.item::<c64>(), val);
-        assert_eq!(array.itemsize(), 8);
+        assert_eq!(array.item::<complex64>(), val);
+        assert_eq!(array.item_size(), 8);
         assert_eq!(array.size(), 1);
         assert!(array.strides().is_empty());
         assert_eq!(array.nbytes(), 8);
@@ -420,7 +422,7 @@ mod tests {
         let array = Array::from_data(&data, &[1], 1);
         assert_eq!(array.data::<i32>(), &data);
         assert_eq!(array.item::<i32>(), 1);
-        assert_eq!(array.itemsize(), 4);
+        assert_eq!(array.item_size(), 4);
         assert_eq!(array.size(), 1);
         assert_eq!(array.strides(), &[1]);
         assert_eq!(array.nbytes(), 4);
@@ -435,7 +437,7 @@ mod tests {
         let data = [1i32, 2, 3, 4, 5];
         let array = Array::from_data(&data, &[5], 1);
         assert_eq!(array.data::<i32>(), &data);
-        assert_eq!(array.itemsize(), 4);
+        assert_eq!(array.item_size(), 4);
         assert_eq!(array.size(), 5);
         assert_eq!(array.strides(), &[1]);
         assert_eq!(array.nbytes(), 20);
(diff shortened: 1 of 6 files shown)


src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+        unsafe { mlx_sys::mlx_array_dim(self.ptr, dim) }
+    }
+
+    /// The array element type.
+    pub fn dtype(&self) -> Dtype {
+        let dtype = unsafe { mlx_sys::mlx_array_get_dtype(self.ptr) };
+        Dtype::try_from(dtype).unwrap()
+    }
+
+    /// Evaluate the array.
```
dcvz, line 320: might leave a comment here about how MLX is lazy and arrays are not fully realized until they are evaluated.

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+
+use crate::sealed::Sealed;
+
+// TODO: camel case?
+// Not using Complex64 because `num_complex::Complex64` is actually Complex<f64>
+#[allow(non_camel_case_types)]
+pub type c64 = Complex<f32>;
+
+/// Array element type
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
dcvz, line 15: for copy, we'd want to make this happen:

```rust
mlx_retain(ctx)
return Array:: from_ptr(ctx)
```

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+        unsafe { mlx_sys::mlx_array_eval(self.ptr) };
+    }
+
+    /// Access the value of a scalar array.
+    pub fn item<T: ArrayElement>(&self) -> T {
+        T::scalar_array_item(self)
+    }
+
+    /// Returns a pointer to the array data
+    pub fn data<T: ArrayElement>(&self) -> &[T] {
```
dcvz, line 332: I don't understand how this is used yet, so I can't comment greatly on what type we need to return here. Shouldn't this be a &[u8] if we're returning byte data? 

either way, let's leave it for now.. but! you want to make sure you call `self.eval()` at the beginning here.

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+    /// The array’s dimension.
+    pub fn ndim(&self) -> usize {
+        unsafe { mlx_sys::mlx_array_ndim(self.ptr) }
+    }
+
+    /// The shape of the array.
+    ///
+    /// Returns: a pointer to the sizes of each dimension.
+    pub fn shape(&self) -> &[i32] {
+        let ndim = self.ndim();
```
dcvz, line 296: Here we also want to verify that ndim is greater than 0 or return empty.

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+        // TODO: check below after ops are implemented
+        // Array must be evaluated, otherwise returns NULL.
+
+        let data = T::array_data(self);
+        let size = self.size();
+        unsafe { std::slice::from_raw_parts(data, size) }
+    }
+}
+
+impl From<bool> for Array {
```
dcvz, line 342: This might be a dumb suggestion but should we introduce a trait for values that can be promoted to an Array? This would work for scalars too. types that conform to ArrayElement would also conform to this new trait. Might not be super Rust-y though, so just a thought.

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+use half::{bf16, f16};
+use mlx_sys::mlx_array;
+use num_complex::Complex;
+
+use crate::sealed::Sealed;
+
+// TODO: camel case?
+// Not using Complex64 because `num_complex::Complex64` is actually Complex<f64>
+#[allow(non_camel_case_types)]
+pub type c64 = Complex<f32>;
```
dcvz, line 12: My feeling is that `complex64` is probably better, just because unlike `i32, u8` or whatever which are ingrained in most devs head, c64 wouldn't be immediately understood.

But with IDEs it doesn't matter much I suppose. c64 is closer to `i32 and u8` let's leave it as such for now.

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+
+    /// New array from existing buffer.
+    ///
+    /// # Parameters
+    ///
+    /// - `data`: A buffer which will be copied.
+    /// - `shape`: Shape of the array.
+    /// - `dim`: Number of dimensions (size of shape).
+    pub fn from_data<T: ArrayElement>(data: &[T], shape: &[i32], dim: i32) -> Self {
+        let ptr = unsafe {
```
dcvz, line 250: We'll want to add a precondition here to validate the shape and data length. something like:

```rust
let total = shape.iter().fold(1, |acc, &x| acc * x);
assert!(total == data.len(), "shape {shape} total {total} != {count} (actual)")
```

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+        assert_eq!(array.size(), 5);
+        assert_eq!(array.strides(), &[1]);
+        assert_eq!(array.nbytes(), 20);
+        assert_eq!(array.ndim(), 1);
+        assert_eq!(array.dim(0), 5);
+        assert_eq!(array.shape(), &[5]);
+        assert_eq!(array.dtype(), Dtype::Int32);
+    }
+
+    // // TODO: fatal runtime error: Rust cannot catch foreign exceptions
```
dcvz, line 448: I think if we start doing the precondition check as I mentioned in another comment, we should no longer get these issues I believe.

dcvz (commented):
It's looking pretty good! Left some comments around. I usually also like to add rust code examples in the docs -- but I know that's cumbersome, so feel free to not do it.

Looking awesome!

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+
+// TODO: camel case?
+// Not using Complex64 because `num_complex::Complex64` is actually Complex<f64>
+#[allow(non_camel_case_types)]
+pub type c64 = Complex<f32>;
+
+/// Array element type
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[repr(u32)]
+pub enum Dtype {
```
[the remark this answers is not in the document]
minghuaw: The organization for now is just following the c header files, but I agree that it would be better to put this in a separate mod. Could be `crate::array::dtype::Dtype`

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+    Float16 = mlx_sys::mlx_array_dtype__MLX_FLOAT16,
+    Float32 = mlx_sys::mlx_array_dtype__MLX_FLOAT32,
+    Bfloat16 = mlx_sys::mlx_array_dtype__MLX_BFLOAT16,
+    Complex64 = mlx_sys::mlx_array_dtype__MLX_COMPLEX64,
+}
+
+impl TryFrom<u32> for Dtype {
+    type Error = u32;
+
+    fn try_from(value: u32) -> Result<Self, Self::Error> {
```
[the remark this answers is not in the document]
minghuaw: `mlx_sys::mlx_array_dtype` is actually just type alias for `u32` (or technically `c_uint`)

src/array.rs
minghuaw: Changed to `complex64`

src/array.rs
minghuaw: That was a mistake. Didn't pay attention to what copilot generated fo the derive. I think we should actually forbid `Clong` or `Copy`. `mlx_retain` just increment the counter but the buffer is not guarded by a `Mutex` or anything similar

dcvz: For Copy this should be ok though right? the counter is incremented and we track the same space in memory?

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+}
+
+impl Array {
+    /// Create a new array from an existing mlx_array pointer.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure the reference count of the array is properly incremented with
+    /// `mlx_sys::mlx_retain`.
+    pub unsafe fn from_ptr(ptr: mlx_array) -> Array {
```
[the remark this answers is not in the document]
minghuaw: This is just following what tch-rs did. Not sure why this is exposed public though. We could just get rid of it completely

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+        Array { ptr }
+    }
+
+    // TODO: should this be unsafe?
+    pub fn as_ptr(&self) -> mlx_array {
```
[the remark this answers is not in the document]
minghuaw: same as `from_ptr`

src/array.rs
minghuaw: I think mlx c++ performs this check already. Need to double check. But panicking on the c++ side would give an fatal error on the rust side, which we will need to deal with at some point

src/array.rs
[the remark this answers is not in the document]
dcvz: We'll need these in the transforms. See #26

src/array.rs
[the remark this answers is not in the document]
dcvz: We'll need these in the transforms. See https://github.com/oxideai/mlx-rs/pull/26

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    fn array_data(array: &Array) -> *const Self {
+        // c64 has the same memory layout as __BindgenComplex<f32>
+        unsafe { mlx_sys::mlx_array_data_complex64(array.ptr) as *const Self }
+    }
+}
+
+pub struct Array {
+    ptr: mlx_array,
```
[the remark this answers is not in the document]
dcvz: That would work, but we'll need this for almost every other type. i.e. device, stream, etc. Having just one name makes it easier rather than specifying for each.  We could just prefix with `c_` though: `c_array`: `c_device` `c_stream`. let's do that.

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    /// Number of elements in the array.
+    pub fn size(&self) -> usize {
+        unsafe { mlx_sys::mlx_array_size(self.ptr) }
+    }
+
+    /// The strides of the array.
+    pub fn strides(&self) -> &[usize] {
+        let ndim = self.ndim();
```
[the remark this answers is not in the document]
minghuaw: I think this is expected behavior. `mlx` follows the `numpy` api, and `strides` and `shape` of a scalar array in `mlx` and `numpy` would be empty.

```python
import mlx.core as mx
import numpy as np

a = mx.array(1)
# mlx python doesn't expose strides somehow
a.shape # would give ()

b = np.array(1)
b.strides # would give ()
b.shape # would give ()
```

src/array.rs
minghuaw: see comments on `strides()`

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+    }
+
+    /// The shape of the array in a particular dimension.
+    ///
+    /// # Panic
+    ///
+    /// Panics if the array is scalar.
+    pub fn dim(&self, dim: i32) -> i32 {
+        // This will panic on a scalar array
+        unsafe { mlx_sys::mlx_array_dim(self.ptr, dim) }
```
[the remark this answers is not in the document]
minghuaw: We should check if this already handled by the c++ impl

src/array.rs
minghuaw: I feel like that should go to the ops, but we could make it a TODO after we implement the ops

src/array.rs
```diff
@@ -0,0 +1,456 @@
...
+
+    /// Evaluate the array.
+    pub fn eval(&mut self) {
+        // This clearly modifies the array, so it should be mutable
+        unsafe { mlx_sys::mlx_array_eval(self.ptr) };
+    }
+
+    /// Access the value of a scalar array.
+    pub fn item<T: ArrayElement>(&self) -> T {
+        T::scalar_array_item(self)
```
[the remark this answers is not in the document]
minghuaw: `mlx` already handles some type conversion, for example the following (should) work

```rust
let a = Array::from_int(1i32);
assert_eq!(a.item::<i16>(), 1i16);
```

But we still have the problem that we currently don't have a way to handle c++ exception. Let's make this a TODO for now.

src/array.rs
minghuaw: This is actually marked as for internal use (or use with caution) in the c++ code, so I was kinda surprised that this is exposed in the c api. 

> Shouldn't this be a &[u8] if we're returning byte data?

The [c++ code](https://github.com/ml-explore/mlx/blob/b18468bf81ded999fda3971289f712920ef0b777/mlx/array.h#L308) returns a `*T`. I think we should probably be consistent with that

> but! you want to make sure you call self.eval() at the beginning here.

That's why I had the TODO there. I want to come back after we have implemented the ops

src/array.rs
minghuaw: Which do you think would be better? A new trait `ArrayElementExt` or just `impl From<T: ArrayElement> for Array`?

src/array.rs
minghuaw: The c++ exceptions are quite nested though, so might be hard to cover all of them. But I agree that this would probably get rid of the problem with c++ exceptions

src/array.rs
dcvz: My thinking was that if we have a new array that can be promoted to array, we could have some of the methods take that trait, and then they could give scalar values without themselves needing to promote it. We could just do that internally. Does that work even if we just do `impl From<T: ArrayElement> for Array`?

src/array.rs
minghuaw: The mlx_array is essentially `Arc<Bytes>`. Maybe wait until we see how many ops mutate the array? If `eval` is the only one that mutate the array, `Clone` and `Copy` is probably fine, because the c++ code checks if the array has been evaluated (if I remember correctly)

src/array.rs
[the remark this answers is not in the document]
minghuaw: Sounds good

src/array.rs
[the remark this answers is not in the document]
minghuaw: Same thing found in the swift binding. Found the code below in the [swift tutorial](https://github.com/ml-explore/mlx-swift/blob/main/Source/Examples/Tutorial.swift)

```swift
// scalars have 0 dimensions
let ndim = x.ndim
assert(ndim == 0)

// scalar shapes are empty arrays
let shape = x.shape
assert(shape == [])
```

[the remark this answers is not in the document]
minghuaw:
```swift
    public var strides: [Int] {
        let ndim = mlx_array_ndim(ctx)
        guard ndim > 0 else { return [] }
        let strides = mlx_array_strides(ctx)!
        return (0 ..< ndim).map { Int(strides[$0]) }
    }
```

src/array.rs
minghuaw: Another potential problem with `Clone` and `Copy`. There is a lot of `std::move` in the ops c++ code, not sure if the underlying `mlx_array` ptr is still valid after those ops

src/array.rs
minghuaw: Should this be renamed to `as_slice`? The swift binding renamed it to `asArray`

minghuaw: Just saw that swift has the `Data` type representing raw bytes. I guess renaming to `as_slice` and `from_slice` would help avoid the confusion

minghuaw: The swift binding calls `eval` internally. We would have to make `as_slice` to take `&mut self` if we want to do the same. Or should we just document this and ask the user to call `eval()`?

src/array.rs
minghuaw: Just realized that this is actually `Clone` and `Copy` on `Dtype`. Let's move the conversation to `Array`

src/array.rs
```diff
@@ -0,0 +1,458 @@
...
+        }
+    }
+
+    fn array_data(array: &Array) -> *const Self {
+        // complex64 has the same memory layout as __BindgenComplex<f32>
+        unsafe { mlx_sys::mlx_array_data_complex64(array.c_array) as *const Self }
+    }
+}
+
+pub struct Array {
```
minghuaw, line 178: I think the swift binding implements `copy()` for array because it doesn't differentiate mutable and immutable borrows. I don't see how we would take advantage of rust's borrow checker if we allow `Clone` and `Copy`

src/array.rs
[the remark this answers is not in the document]
minghuaw: > we should add a check here to make sure ndim is > 0 otherwise let's return empty.

I misread. This is already returning an empty slice if `ndim == 0`. Check the scalar array unit test

src/array.rs
minghuaw: I am not sure if I understand this correctly. Is the example below depicting what you were thinking?

```rust
let arr = Array::from_int(1);
let result: Array = add(arr, 1); // 1i32 can be converted into an `Array`
```

src/array.rs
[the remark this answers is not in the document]
minghuaw: I think maybe we would want to wait until `as_type` is implemented in the ops. Using `mlx_sys::mlx_astype` gives an linker error.

src/array.rs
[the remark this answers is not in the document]
minghuaw: Moved `Dtype` to `crate::dtype::Dtype`

src/array.rs
minghuaw: It's already returning an empty slice. See scalar array unit test

src/array.rs
dcvz: Nah, I think we should assume intent. My guess is that it'll be easier to forget that you need to call eval rather than doing it. We can always change it later.

src/array.rs
dcvz: Yes, this was my idea. But we can also leave it for now, and adjust as we actually start adding ops and start to get a feel for where the API could feel better

src/array.rs
dcvz: I'd say where we can catch them on the Rust side, let's just do it

src/array.rs
minghuaw: I think `From<T>` would suffice. We would just need to do something like this when we impl ops

```rust
fn add(arr1: impl Into<Array>, arr2: impl Into<Array>, s: StreamOrDevice) -> Array {
    let arr1 = arr1.into();
    let arr2 = arr2.into();
    let c_array = unsafe { mlx_add(arr1.c_array, arr2.c_array, s) };
    Array { c_array }
}
```

src/array.rs
minghuaw: This function has been renamed to `from_slice`, which now checks whether the shape and data length match

src/array.rs
minghuaw: My concern is still data racing. Let me double check if the c++ side does anything in this regard for `eval()`

src/array.rs
minghuaw: Perform checks in rust to avoid c++ exception

src/array.rs
minghuaw: The way mlx determines whether an `array` is `eval`ed is by checking whether the `array_desc_->data` ptr is set. The `array_desc_->data` ptr is a `std::shared_ptr` but not protected by `atomic`. Plus the usual impl of the primitives seem to allocate and set the `data` ptr before all output are written to the newly allocated buffer. So if we call `mlx_array_eval` implicitly within a fn that only immutably borrows, we could likely end up with data racing

minghuaw: `as_slice` returns `None` if the array is not `eval`ed.

PR #26, opened 2024-04-17
feat(api): adding transforms

dcvz:
Further supports #25


minghuaw:
Was this also closed because you didn't want it to be "owned" by you? @dcvz

minghuaw:
Re-opening to continue work on transforms

minghuaw:
The remaining transforms APIs seem complicated and should probably be attempted in separate PRs

minghuaw:
I think we are only missing the compile ops for the transforms ops with this PR. `async_eval` is included in a separate PR

minghuaw:
Added compile. Because of the lack of function overloading, only the most general version of the `compile` (https://swiftpackageindex.com/ml-explore/mlx-swift/main/documentation/mlx/compile(inputs:outputs:shapeless:_:)-7korq) is added

minghuaw:
Let's try to get this merged first and then work on the ergonomics in a separate PR.

src/transforms/mod.rs
```diff
@@ -0,0 +1,396 @@
...
+    argument_numbers: &'a [i32],
+) -> impl FnMut(&[Array]) -> Result<(Vec<Array>, Vec<Array>), Exception> + 'a
+where
+    F: FnMut(&[Array]) -> Vec<Array> + 'a,
+{
+    build_value_and_gradient(f, argument_numbers)
+}
+
+/// Similar to `value_and_grad`, but unwraps the result.
+pub fn value_and_grad_unwrapped<'a, F>(
```
dcvz, line 230: we haven't really done `unwrapped` so far, is this specifically necessary for something? Otherwise I'd wait just a bit before adding them just here. Same for the other instances of `unwrapped` in this file.

src/utils.rs
```diff
@@ -238,3 +238,131 @@ where
...
+{
+    unsafe {
+        let raw_closure: *mut F = payload as *mut _;
+        // Let the box take care of freeing the closure
+        let mut closure = Box::from_raw(raw_closure);
+        let arrays = mlx_vector_array_values(vector_array);
+        let result = closure(&arrays);
+        // We should probably keep using new_mlx_vector_array here instead of VectorArray
+        // since we probably don't want to drop the arrays in the closure
+        new_mlx_vector_array(result)
```
dcvz, line 333: why would things get dropped if we use VectorArray?

minghuaw: Because we have implemented Drop for VectorArray right?

src/transforms/mod.rs
minghuaw: It was originally added so that one could nest `grad()` to compute 2nd order gradient, but there's still lifetime issue. So I guess we could remove them.

minghuaw: Resolved in [1083a60](https://github.com/oxideai/mlx-rs/pull/26/commits/1083a60eb69b8c2dc37ae125bd90b0d65627d6c8)

PR #28, opened 2024-04-17
feat(api): Add Stream and StreamOrDevice

dcvz:


src/stream.rs
```diff
@@ -0,0 +1,104 @@
...
+pub struct Stream {
+    c_stream: mlx_sys::mlx_stream,
+}
+
+impl Stream {
+    fn new_with_mlx_mlx_stream(stream: mlx_sys::mlx_stream) -> Stream {
+        Stream { c_stream: stream }
+    }
+
+    pub fn new() -> Stream {
```
minghuaw, line 74: I think `clippy` would suggest impl the `Default` trait while keeping the `new()` method

src/stream.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    /// The `[Stream::default_stream()] on the [Device::gpu()]
+    pub fn gpu() -> StreamOrDevice {
+        StreamOrDevice {
+            stream: Stream::default_stream(&Device::gpu()),
+        }
+    }
+
+    #[allow(dead_code)]
+    fn get_ctx(&self) -> mlx_sys::mlx_stream {
```
minghuaw, line 51: Maybe also name it `as_ptr`? Or we can rename `Array::as_ptr`. Just thinking we should be consistent in naming these methods that expose the inner c ptr

src/device.rs
```diff
@@ -9,13 +9,13 @@ pub enum DeviceType {
 /// Representation of a Device in MLX.
 #[derive(Debug)]
 pub struct Device {
-    ctx: mlx_sys::mlx_device,
+    pub(crate) c_device: mlx_sys::mlx_device,
 }
 
 impl Device {
     pub fn new_default() -> Device {
```
minghuaw, line 16: impl the `Default` trait instead?

src/stream.rs
dcvz: Yeah, agree on consistency. Let's use `as_ptr()` -- usually I'm more in favor of exposing the inner `c_xx` but here its actually the subproperty that has it.

src/stream.rs
```diff
@@ -0,0 +1,112 @@
...
+///
+/// If omitted it will use the [default()], which will be [Device::gpu()] unless
+/// set otherwise.
+pub struct StreamOrDevice {
+    stream: Stream,
+}
+
+impl StreamOrDevice {
+    #[allow(dead_code)]
+    fn new(stream: Stream) -> StreamOrDevice {
```
minghuaw, line 16: Maybe make this public given that there is already a pub fn for device

src/device.rs
```diff
@@ -25,7 +25,7 @@ impl Device {
         };
```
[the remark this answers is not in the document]
dcvz: Done! Added `num_enum` for derived `into` and `try_from` -- updated Dtype as well.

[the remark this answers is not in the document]
minghuaw: I have been doing these conversion manually lol. Should have used `num_enum` in the first place

src/stream.rs
minghuaw: Should it be public though?

dcvz: We can get rid of it while we have no uses. I'll delete.

minghuaw (approved):

PR #32, opened 2024-04-18
feat(api): Add factory (creation) ops

dcvz:


minghuaw:
I have some general question regarding the `ops` mod that I think we should think through before diving deep into these functions

1. Should the be associated functions (ie. `Array::zeros()`) or just regular functions (ie. `crate::ops::zeros()`), or both?
2. Should we just forget about `zeros()` and `zeros_device()` for now? Maybe just do `zeros()` and provide a macro that allow variable number of arguments later?

src/ops.rs
```diff
@@ -0,0 +1,411 @@
...
+    /// ```rust
+    /// use mlx::{array::Array, stream::StreamOrDevice};
+    /// Array::zeros::<f32>(&[5, 10], StreamOrDevice::default());
+    /// ```
+    ///
+    /// # Parameters:
+    /// - shape: Desired shape
+    /// - stream: Stream or device to evaluate on
+    pub fn zeros<T: ArrayElement>(shape: &[usize], stream: StreamOrDevice) -> Array {
+        let shape = shape.iter().map(|x| *x as i32).collect::<Vec<i32>>();
```
minghuaw, line 19: We should check for overflow when converting from `usize` to `i32`

minghuaw: Or just make it `&[i32]` to avoid additional memory allocation

src/ops.rs
```diff
@@ -0,0 +1,411 @@
+use crate::array::ArrayElement;
+use crate::{array::Array, stream::StreamOrDevice};
+use num_traits::NumCast;
+
+impl Array {
+    /// Construct an array of zeros.
+    ///
+    /// Example:
```
minghuaw, line 8: I think the rust convention for example usually has # before Example

```rust
/// # Example
/// ```rust
/// foo();
/// ```
fn foo() { }

```

src/ops.rs
```diff
@@ -0,0 +1,411 @@
...
+                T::DTYPE.into(),
+                ctx,
+            ))
+        }
+    }
+
+    /// Construct an array with the given value.
+    ///
+    /// Constructs an array of size `shape` filled with `values`. If `values`
+    /// is an :obj:`array` it must be <doc:broadcasting> to the given `shape`.
```
minghuaw, line 95: Fix non-rust doc items like ":obj: `array`", which should be [`Array`]

src/ops.rs
```diff
@@ -0,0 +1,411 @@
...
+                ctx,
+            ))
+        }
+    }
+
+    /// Element-wise absolute value.
+    ///
+    /// # Parameters:
+    /// - stream: stream or device to evaluate on
+    pub fn abs(&self, stream: StreamOrDevice) -> Array {
```
minghuaw, line 262: Not specific to `abs()`, but for other functions, we should probably test whether it should take reference or ownership, since a lot of the ops impl moves the pointer with `std::move`.

src/ops.rs
```diff
@@ -0,0 +1,411 @@
...
+
+        array.eval();
+        let data: &[f32] = array.as_slice().unwrap();
+        assert_eq!(data, &[1.0, 0.0, 0.0, 1.0, 1.0, 0.0, 1.0, 1.0, 1.0]);
+    }
+
+    #[test]
+    fn test_abs() {
+        let data = [1i32, 2, -3, -4, -5];
+        let array = Array::from_slice(&data, &[5]);
```
minghuaw, line 404: Maybe we can add a test to see if the original array is still valid after evaluating `result`

dcvz:
> I have some general question regarding the `ops` mod that I think we should think through before diving deep into these functions
> 
> 
> 
> 1. Do we want to impl all `ops` and merge altogether or do it incrementally?
> 
> 2. Should the be associated functions (ie. `Array::zeros()`) or just regular functions (ie. `crate::ops::zeros()`), or both? I think the debate here is that `mlx` itself doesn't make ops associated functions; however, rust primitives like `i32` make functions like `add`, `sub` associated functions (well, technically through trait)
> 
> 3. Should we just forget about `zeros()` and `zeros_device()` for now? Maybe just do `zeros()` and provide a macro that allow variable number of arguments later?
> 
> 4. Should we make three versions of each function? Using `zeros` as an example
> 
>     a. `fn zeros_unchecked() -> Array` - does not check the parameter at all. This might have to be `unsafe` so that the user should use it with caution
> 
>     b. `fn try_zeros() -> Result<Array, Error>` - checks parameters, returns an error if parameters are not valid
> 
>     c. `fn zeros() -> Array` - checks parameters, panics if not valid
> 
>     The implementation would be like follows. `zeros_unchecked` simply wraps around the c binding; then `try_zeros` checks parameter and call `zeros_unchecked`; then `zeros` calls `try_zeros` but unwrap the result

1. I'm thinking of doing the ops in parts. This PR is for the factories then I'd make an issue listing the missing ones and we can tackle those in segments.

2. I think the Rust community is used to associated functions. Plus it's what most other languages python / swift do.

3. I was working on adding the device variant yesterday. So I'll still do that for these.

4. I need to think some more about this suggestion and try it out.. it feels like we're still trying to mitigate a lot of the class of issues a typed Array/Tensor API would get rid of and I continue being torn about the direction we're taking. I'll also check out torch-rs and other libs to see how they handle this.

minghuaw:
> 2. ... Plus it's what most other languages python / swift do.

I disagree on this though. `numpy` and `mlx` in python clearly doesn't do associated functions. For example

```python
import numpy as np
import mlx.core as mx

a = np.zeros(5)
b = mx.zeros(5)
```

> 4. it feels like we're still trying to mitigate a lot of the class of issues a typed Array/Tensor API would get rid of and I continue being torn about the direction we're taking. I'll also check out torch-rs and other libs to see how they handle this.

I think this feels complicated simply because handling ffi and linear algebra ops are inherently complicated. 

The idea about the three variants comes from that `try_*`/`*_checked`, and `_unchecked` are like the typical way of differentiating the hidden cost of assertions. For example, `slice` have three variants of `split_at` - [`split_at`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at), [`split_at_checked`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at_checked), and [`split_at_unchecked`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at_unchecked)

A typed array is not free of some of these problems either. One use case for these machine learning library is to load a model from somewhere else and run that model. You don't have any information regarding the type or shape about the tensor or operations at compile time at all. So the dynamic array is something that we have to implement anyway, and these problems won't just disappear.

src/ops.rs
dcvz: Did that! Makes more sense

src/ops.rs
dcvz: Did!

src/ops/factory.rs
```diff
@@ -0,0 +1,901 @@
...
+    ///
+    /// - start: start value
+    /// - stop: stop value
+    /// - count: number of samples
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn linspace_device<T, U>(
+        start: U,
+        stop: U,
+        count: Option<i32>,
```
minghuaw, line 421: Maybe the default count should be documented?

src/ops/mod.rs
```diff
@@ -0,0 +1,4 @@
+mod array;
+mod factory;
+
+pub use array::*;
```
minghuaw: I think this line is not needed. The `array` mod and `factory` mod simply provide impls, and all those pub associated functions are already exposed to the user

src/ops/factory.rs
```diff
@@ -0,0 +1,901 @@
...
+    /// use mlx::{Array, StreamOrDevice};
+    /// unsafe { Array::zeros_device_unchecked::<f32>(&[5, 10], StreamOrDevice::default()) };
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - shape: Desired shape
+    /// - stream: Stream or device to evaluate on
+    #[default_device]
+    pub unsafe fn zeros_device_unchecked<T: ArrayElement>(
```
minghuaw, line 41: `clippy` seems to complain about not documenting why this is unsafe. It wants to see something like. But this is not a big deal.

```rust
/// # Safety
/// 
/// <placeholder>
```

minghuaw (approved):
I've made some minor changes

1. Corrected my name in `mlx-macros/Cargo.toml` 
2. Changed visibility of `c_array` in `Array` to `pub(crate)`
3. Changed `eye_device_unsafe` to `eye_device_unchecked`
4. Removed the unnecessary `pub use array::*` in `src/ops/mod.rs`

I think this is good to go. `clippy` complained about not having a `/// # Safety` section in the documentations for the unsafe functions, but I think this is not a big deal given that folks should understand this is a binding crate. We could also just make a general note about this at the crate level documentation and turn off this `clippy` warning later.

src/ops/factory.rs
minghuaw: This is not a big deal tho, since the user can always refer to the mlx python docs. It's up to you whether this should be documented here

PR #36, opened 2024-04-20
Add `try_*` and `*_unchecked` variant to `as_slice`

minghuaw:


src/array.rs
```diff
@@ -317,18 +316,92 @@ impl Array {
         T::scalar_array_item(self)
     }
 
-    /// Returns a pointer to the array data
+    // /// Returns a pointer to the array data
+    // ///
+    // /// Returns `None` if the array is not evaluated.
+    // pub fn as_slice<T: ArrayElement>(&self) -> Option<&[T]> {
```
dcvz, line 322: We can get rid of this now, right?

dcvz (approved):
Looks good! Just one small place where we can remove code

src/array.rs
minghuaw: yeah, forgot to do that

PR #39, opened 2024-04-20
feat(api): Arithmetic & Logical ops

dcvz:


src/ops/logical.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// # Example
+    ///
+    /// ```rust
+    /// use mlx::Array;
+    /// let a: Array = false.into();
+    /// let mut b = a.logical_not_device(Default::default());
+    ///
+    /// b.eval();
+    /// let b_data: &[bool] = b.as_slice();
+    /// // b_data == [true]
```
minghuaw, line 17: Maybe do `assert_eq!()` here so that cargo test also checks for correctness?

src/ops/logical.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// let c_data: &[bool] = c.as_slice();
+    /// // c_data == [true, true, true]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to compare
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn eq_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 50: What if two arrays are of different shape? What about doing the `*_unchecked` and `try_*` here?

minghuaw: The c++ code checks whether the two shapes are broadcastable

src/ops/logical.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// let c_data: &[bool] = c.as_slice();
+    /// // c_data == [true, true, true]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to compare
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn le_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 82: Same problem as `eq_device`

src/ops/logical.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// let c_data: &[bool] = c.as_slice();
+    /// // c_data == [true, true, true]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to compare
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn ge_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 114: Same problem as `eq_device`

src/ops/logical.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// let c_data: &[bool] = c.as_slice();
+    /// // c_data == [false, false, false]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to compare
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn ne_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 146: Same problem as `eq_device`

src/ops/logical.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// let c_data: &[bool] = c.as_slice();
+    /// // c_data == [false, false, false]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to compare
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn lt_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 178: Same problem as `eq_device`

src/ops/logical.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// let c_data: &[bool] = c.as_slice();
+    /// // c_data == [false, false, false]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to compare
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn gt_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 210: Same problem as `eq_device`

src/ops/logical.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// let c_data: &[bool] = c.as_slice();
+    /// // c_data == [true, false, false]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to compare
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn logical_and_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 242: We probably want to check not only the shapes and also the dtype as well

src/ops/logical.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// let c_data: &[bool] = c.as_slice();
+    /// // c_data == [true, true, true]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to compare
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn logical_or_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 274: same as `logical_and_device`

src/ops/logical.rs
```diff
@@ -0,0 +1,554 @@
...
+    ///
+    /// # Params
+    ///
+    /// - other: array to compare
+    /// - rtol: relative tolerance = defaults to 1e-5 when None
+    /// - atol: absolute tolerance - defaults to 1e-8 when None
+    /// - equal_nan: whether to consider NaNs equal -- default is false when None
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn all_close_device(
```
minghuaw, line 314: same as `logical_and_device`

src/ops/logical.rs
```diff
@@ -0,0 +1,554 @@
...
+    /// // c == [true]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to compare
+    /// - equal_nan: whether to consider NaNs equal -- default is false when None
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn array_eq_device(
```
minghuaw, line 357: same as `logical_and_device`

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    /// # Example
+    ///
+    /// ```rust
+    /// use mlx::Array;
+    /// let array = Array::from_slice(&[1i32, 2, -3, -4, -5], &[5]);
+    /// let mut result = array.abs();
+    ///
+    /// result.eval();
+    /// let data: &[i32] = result.as_slice();
+    /// // data == [1, 2, 3, 4, 5]
```
minghuaw, line 17: Maybe use `assert_eq!()` in the doc example?

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    /// c.eval();
+    /// let c_data: &[f32] = c.as_slice();
+    /// // c_data == [5.0, 7.0, 9.0]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to add
+    /// - stream: stream or device to evaluate on
+    pub fn add_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 50: What if the shapes of the two arrays are not broadcastable? We probably also want to check combination of two `dtype`s for binary ops that are supported

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    /// c.eval();
+    /// let c_data: &[f32] = c.as_slice();
+    /// // c_data == [-3.0, -3.0, -3.0]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to subtract
+    /// - stream: stream or device to evaluate on
+    pub fn sub_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 81: same problem as `add_device`

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    ///
+    /// b.eval();
+    /// let b_data: &[f32] = b.as_slice();
+    /// // b_data == [-1.0, -2.0, -3.0]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - stream: stream or device to evaluate on
+    pub fn neg_device(&self, stream: StreamOrDevice) -> Array {
```
minghuaw, line 110: same problem as `add_device`

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    /// use mlx::Array;
+    /// let a = Array::from_slice(&[1.0, 2.0, 3.0], &[3]);
+    /// let b = Array::from_slice(&[4.0, 5.0, 6.0], &[3]);
+    /// let mut c = a.mul_device(&b, Default::default());
+    ///
+    /// c.eval();
+    /// let c_data: &[f32] = c.as_slice();
+    /// // c_data == [4.0, 10.0, 18.0]
+    /// ```
+    pub fn mul_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 130: same problem as `add_device`

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    /// c.eval();
+    /// let c_data: &[f32] = c.as_slice();
+    /// // c_data == [0.25, 0.4, 0.5]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to divide
+    /// - stream: stream or device to evaluate on
+    pub fn div_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 161: same problem as `add_device`

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    /// c.eval();
+    /// let c_data: &[f32] = c.as_slice();
+    /// // c_data == [1.0, 8.0, 81.0]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to raise to the power of
+    /// - stream: stream or device to evaluate on
+    pub fn pow_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 192: same problem as `add_device`

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    /// c.eval();
+    /// let c_data: &[f32] = c.as_slice();
+    /// // c_data == [1.0, 3.0, 2.0]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to divide
+    /// - stream: stream or device to evaluate on
+    pub fn rem_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 223: same problem as `add_device`

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    /// use mlx::Array;
+    /// let a = Array::from_slice(&[1.0, 4.0, 9.0], &[3]);
+    /// let mut b = a.sqrt_device(Default::default());
+    ///
+    /// b.eval();
+    /// let b_data: &[f32] = b.as_slice();
+    /// // b_data == [1.0, 2.0, 3.0]
+    /// ```
+    #[default_device]
+    pub fn sqrt_device(&self, stream: StreamOrDevice) -> Array {
```
minghuaw, line 247: same problem as `add_device`

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    /// use mlx::Array;
+    /// let a = Array::from_slice(&[0.0, 1.0, 2.0], &[3]);
+    /// let mut b = a.cos_device(Default::default());
+    ///
+    /// b.eval();
+    /// let b_data: &[f32] = b.as_slice();
+    /// // b_data == [1.0, 0.54030234, -0.41614687]
+    /// ```
+    #[default_device]
+    pub fn cos_device(&self, stream: StreamOrDevice) -> Array {
```
minghuaw, line 265: same problem as `add_device`

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    /// let c_data: &[f32] = c.as_slice();
+    /// // c_data == [0.25, 0.4, 0.5]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to divide
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn floor_divide_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 331: same problem as `add_device`

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    /// // produces a [2, 3] result
+    /// let mut c = a.matmul_device(&b, Default::default());
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - other: array to multiply
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn matmul_device(&self, other: &Array, stream: StreamOrDevice) -> Array {
```
minghuaw, line 443: same problem as `add_device`

minghuaw: we would probably need a separate fn to check shapes for matmul

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,875 @@
...
+    }
+
+    /// Round to the given number of decimals.
+    ///
+    /// # Params
+    ///
+    /// - decimals: number of decimals to round to - default is 0 if not provided
+    /// - stream: stream or device to evaluate on
+    #[default_device]
+    pub fn round_device(&self, decimals: impl Into<Option<i32>>, stream: StreamOrDevice) -> Array {
```
minghuaw, line 478: What happens on the c++ side if `decimals < 0`? Does it throw an exception?

minghuaw: I think this is fine as long as the c++ side doesn't just throw and exception. Maybe we should test that

src/array.rs
```diff
@@ -137,6 +138,55 @@ pub struct Array {
     pub(crate) c_array: mlx_array,
 }
```
minghuaw, line 140: Should we also impl `PartialEq` with `mlx_equal` and `PartialOrd` with the `ge`, `le`, `gt`, and `lt` functions?

minghuaw (commented):
We should probably take `mlx`'s code that checks if two shapes are broadcastable into `utils`. I feel like we are going to use that a lot

src/array.rs
dcvz: I really considered it... but the way these work in `mlx` is that they return arrays where each element has been compared -- AFAIU in Rust these traits return a `bool` which means that we'd change the semantics a way to make each item is in fact equal, bigger, smaller whatever.

`mlx` does have an `array_equal` which does return an array with a single bool value in it which we could maybe use, but not sure if its clear that its using this comparison. Open to opinions here.

src/ops/arithmetic.rs
dcvz: turns out this is valid in mlx, it just returns an array of zeros

src/error.rs
```diff
@@ -1,12 +1,37 @@
 use thiserror::Error;
 
-#[derive(Error, Debug)]
+#[derive(Error, PartialEq, Debug)]
+pub enum MLXError {
```
minghuaw, line 4: This should be `MlxError` following the rust naming convention

src/error.rs
```diff
@@ -1,12 +1,37 @@
...
+    #[error("data store error: {0}")]
+    DataStoreError(#[from] DataStoreError),
```
minghuaw, line 6: `clippy` will complain that all variants end with `Error`. Maybe just to

```rust
pub enum MlxError {
    DataStore(#[from] DataStoreError),
    Operation(#[from] OperationError),
    ...
}
```

src/error.rs
```diff
@@ -19,6 +44,6 @@ pub enum AsSliceError {
     Null,
 
     /// The output dtype does not match the data type of the array.
-    #[error("Desired output dtype does not match the data type of the array.")]
-    DtypeMismatch,
+    #[error("dtype mismatch: {0}")]
+    DtypeMismatch(String),
```
minghuaw, line 48: I generally try to avoid carrying a runtime string as much as possible. The same error message can be obtained by just carrying the desired `Dtype` and actual `Dtype`

minghuaw: I think some of the other errors types don't really need to carry a `String` either

mlx-macros/src/lib.rs
```diff
@@ -51,3 +51,51 @@ pub fn default_device(_attr: TokenStream, item: TokenStream) -> TokenStream {
 
     TokenStream::from(expanded)
 }
+
+#[proc_macro_derive(GenerateDTypeTestCases)]
```
minghuaw, line 55: Minor thing, but I think it should be `GenerateDtypeTestCases` if we were to follow the rust naming convention strictly

src/utils.rs
```diff
@@ -17,3 +19,145 @@ pub(crate) fn mlx_describe(ptr: *mut ::std::os::raw::c_void) -> Option<String> {
...
+impl Array {
+    /// Helper method to check if an array can be reshaped to a given shape.
+    pub fn can_reshape_to(&self, shape: &[i32]) -> bool {
+        if self.shape() == shape {
+            return true;
+        }
+
+        let mut size = 1;
+        let mut infer_idx: isize = -1;
+        for i in 0..shape.len() {
```
minghuaw, line 50: Just curious when should a shape carry negative element

src/ops/arithmetic.rs
```diff
@@ -0,0 +1,1698 @@
...
+        stream: StreamOrDevice,
+    ) -> Result<Array, OperationError> {
+        if self.ndim() == 0 || other.ndim() == 0 {
+            return Err(OperationError::WrongInput(
+                "Got 0 dimension input. Inputs must have at least one dimension.".to_string(),
+            )
+            .into());
+        }
+
+        // if we have a 1D array, it will be reshaped to 2D
```
minghuaw, line 1117: Maybe we could do `[i32; 2]` instead to avoid heap alloc for something that has only 2 element?

src/ops/arithmetic.rs
dcvz: We decided not to, after you looked into it

src/error.rs
dcvz: You made this one :D I just updated the message to be similar to the others

src/ops/arithmetic.rs
dcvz: problem is other side is `&[i32]` didn't know how to make these two types play with each other but I've now gone with a different approach that avoids the heap alloc

src/ops/logical.rs
dcvz: We decided not to do this after investigation

src/ops/logical.rs
dcvz: I've added a check for broadcasting but dtypes don't need to be compared -- mlx works on any two types, it always converts them to bool first.

src/ops/arithmetic.rs
dcvz: mlx doesn't have such check, it promotes types and checks for broadcast

src/ops/arithmetic.rs
dcvz: The `try_matmul_device` has checks now

src/utils.rs
dcvz: good question, just ported this over from mlx

src/error.rs
dcvz: I think its useful for actually giving context on what params may have been wrong and what we received. It should help a lot with debugging. Check out how the errors have been used so far.

src/ops/reduction.rs
```diff
@@ -0,0 +1,298 @@
...
+    /// // results == [false]
+    /// ```
+    ///
+    /// # Params
+    ///
+    /// - keep_dims: Whether to keep the reduced dimensions -- defaults to false if not provided
+    /// - stream: The stream to execute the operation on
+    #[default_device]
+    pub fn all_device(&self, keep_dims: impl Into<Option<bool>>, stream: StreamOrDevice) -> Array {
+        let axes: Vec<i32> = (0..self.ndim() as i32).collect();
```
dcvz, line 239: @minghuaw know a better way to do this one?

src/utils.rs
minghuaw: I wonder what is negative shape intended for? Is -1 the only possible negative value?

src/error.rs
minghuaw: A potential alternative is 

```rust
pub enum AsSliceError {
    Null,
    DtypeMismatch {
        expecting: Dtype,
        found: Dtype,
    }
}
```

But let's leave it as it is for now and come back later

src/utils.rs
dcvz: resolved on discord + added one more test for dimensions other than -1

src/error.rs
dcvz: I went ahead and changed it!

src/ops/logical.rs
```diff
@@ -0,0 +1,1323 @@
...
+    #[default_device]
+    pub unsafe fn all_close_device_unchecked(
+        &self,
+        other: &Array,
+        rtol: impl Into<Option<f64>>,
+        atol: impl Into<Option<f64>>,
+        equal_nan: impl Into<Option<bool>>,
+        stream: StreamOrDevice,
+    ) -> Array {
+        let is_close = self.is_close_device_unchecked(other, rtol, atol, equal_nan, stream.clone());
```
minghuaw, line 868: Why not just use `mlx_allclose`? Though these should be the same

src/ops/reduction.rs
```diff
@@ -0,0 +1,298 @@
...
+    ///
+    /// # Params
+    ///
+    /// - axes: The axes to reduce over
+    /// - keep_dims: Whether to keep the reduced dimensions -- defaults to false if not provided
+    /// - stream: The stream to execute the operation on
+    #[default_device]
+    pub fn all_axes_device(
+        &self,
+        axes: &[i32],
```
minghuaw, line 29: The python api allows `axes` to be optional. Should this be optional too?

src/ops/logical.rs
dcvz: I guess for `unchecked` I could just use `mlx` directly. It shouldn't matter, but I can do that instead.

src/ops/reduction.rs
minghuaw: Should we just merge `all_axes_*`, `all_axis_*` into `all_*`? Since that's how we are dealing with optional params in other methods?

src/ops/reduction.rs
dcvz: python allows calling `all` with either an int, a vec of ints or nothing -- in Rust I've split these into three methods: `all_axes (takes array)`, `all_axis (takes one)` or `all` which takes none.

I'm not sure there's a good way in Rust to denote these three different scenarios in one.

minghuaw: I think in mlx c++ code the integer axis is wrapped into `std::vector<int>` and then call the array axes variant. So maybe `fn all(&self, axes: Option<&[i32]>, keep_dims: Option<bool>, s: StreamOrDevice)` is fine? The user would just need to wrap a single axis into a single element slice themselves?

dcvz: Fine with me! Makes things simpler. Will make the change.

minghuaw (approved):
I think it's good to go

PR #40, opened 2024-04-21
feat(api): impl Clone and Debug for Device and Stream

minghuaw:
1. Added `Clone` impl for `Device`, `Stream`, and `StreamOrDevice`
2. Deny `missing_debug_implementations` (pub types that don't have `Debug` impl will give an compile error`)

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -1,7 +1,9 @@
+use mlx_sys::mlx_retain;
+
 use crate::utils::mlx_describe;
 
 ///Type of device.
-#[derive(num_enum::IntoPrimitive)]
+#[derive(num_enum::IntoPrimitive, Clone, Copy)]
 #[repr(u32)]
 pub enum DeviceType {
     Cpu = mlx_sys::mlx_device_type__MLX_CPU,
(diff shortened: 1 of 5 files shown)


src/device.rs
```diff
@@ -55,6 +85,15 @@ impl Default for Device {
     }
 }
 
+impl std::fmt::Debug for Device {
```
dcvz, line 88: Can we not just use the display impl here?

src/stream.rs
```diff
@@ -102,6 +133,14 @@ impl Default for Stream {
     }
 }
 
+impl std::fmt::Debug for Stream {
```
dcvz, line 136: Same as debug for device

dcvz (approved):
LGTM 👍 some small comments

src/device.rs
minghuaw: I think this is more a rust convention thing. I read somewhere that a recommended practice for library crates is to always provide `Debug` impl, which is why I added the `#![deny(missing_debug_implementations)]`. And `Display` is considered rather optional, but it doesn't hurt to provide both

minghuaw: My brain's rusted lol. Misread your comment. Yeah, we could, copilot just generated the same code and I was too lazy to change that

PR #44, opened 2024-04-23
feat(api): impl ops/conversion

minghuaw:


minghuaw:
1. Added `Array::as_type_device` impl
2. Bug fix `Array::try_as_slice` didn't check if data pointer was null

src/ops/conversion.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// use mlx::Array;
+    ///
+    /// let array = Array::from_slice(&[1i16,2,3], &[3]);
+    /// let mut new_array = array.as_type::<f32>();
+    /// new_array.eval();
+    ///
+    /// assert_eq!(new_array.dtype(), mlx::Dtype::Float32);
+    /// assert_eq!(new_array.shape(), &[3]);
+    /// assert_eq!(new_array.item_size(), 4);
+    /// assert_eq!(new_array.as_slice::<f32>(), &[1.0,2.0,3.0]);
```
dcvz, line 20: are we doing asserts in the examples after all?

src/ops/conversion.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// ```
+    #[default_device]
+    pub fn as_type_device<T: ArrayElement>(&self, stream: StreamOrDevice) -> Array {
+        unsafe {
+            let new_array = mlx_sys::mlx_astype(self.c_array, T::DTYPE as u32, stream.as_ptr());
```
dcvz, line 25:
```suggestion
            let new_array = mlx_sys::mlx_astype(self.c_array, T::DTYPE.into(), stream.as_ptr());
```

src/ops/conversion.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    macro_rules! test_as_type {
+        ($src_type:ty, $src_val:expr, $dst_type:ty, $dst_val:expr, $len:expr) => {
+            paste::paste! {
+                #[test]
+                fn [<test_as_type_ $src_type _ $dst_type>]() {
+                    let array = Array::from_slice(&[$src_val; $len], &[$len as i32]);
+                    let s = StreamOrDevice::default();
+
+                    let mut new_array = array.as_type_device::<$dst_type>(s);
```
dcvz, line 47: not that it matters much but you should be able to leverage the default device method here. `as_type`

dcvz (approved):
LGTM 👍 left some small comments but feel free to address those you feel like and merge

src/ops/conversion.rs
minghuaw: I don't think we have to. I just felt for this function it's a better way to show what remains the same what's changed in the type conversion

dcvz: Makes sense!

src/ops/conversion.rs
minghuaw: Yeah, this is mostly generated by copilot, so I didn't bother changing that. But I agree that testing`as_type` is better here

dcvz (approved):

PR #45, opened 2024-04-23
Fixed a bug where `try_as_slice` doesn't check if data ptr is null

minghuaw:


src/ops/conversion.rs
minghuaw, line 1: The bug was found while implementing the conversion ops, and I forgot to remove the conversion mod files from the commit

minghuaw:
Just realized the the conversion mod really just need one function. Will close this an submit the change along with the conversion PR #44

PR #46, opened 2024-04-24
feat(api): cumulative ops

dcvz:


src/utils.rs
```diff
@@ -97,6 +97,20 @@ impl Array {
 
         return true;
     }
+
+    /// Helper method to validate an axis is in bounds.
+    pub fn validate_axis_in_bounds(&self, axis: Option<i32>) -> Result<(), OperationError> {
```
minghuaw, line 102: Maybe make this more generic. I think we will have more functions using this in the future. Something like

```rust
fn is_axis_in_bound(axis: Option<i32>, ndim: usize) -> bool;
```

minghuaw: Or

```rust
fn validate_axis_in_bound(axis: Option<i32>, ndim: usize) -> Result<(), (i32, usize)>;
```

PR #47, opened 2024-04-24
feat(api): convolution ops

dcvz:


src/ops/convolution.rs
```diff
@@ -0,0 +1,495 @@
...
+            groups,
+            flip,
+            stream.as_ptr(),
+        ))
+    }
+}
+
+/// General convolution over an input with several channels.
+///
+/// > Only 1d and 2d convolutions are supported at the moment
```
minghuaw, line 41: This put a tab on the generate doc, which is not common in the rust community. I'd prefer to avoid it if possible, but it's up to you.

src/ops/convolution.rs
```diff
@@ -0,0 +1,495 @@
...
+///
+/// > the default `groups: 1` is currently supported
+///
+/// # Params
+/// - array: Input array of shape `&[N, ..., C_in]`
+/// - weight: Weight array of shape `&[C_out, ..., C_in]`
+/// - strides: The kernel strides. All dimensions get the same stride if only one number is specified.
+/// - padding: The input padding. All dimensions get the same padding if only one number is specified.
+/// - kernel_dilation: `The  kernel dilation. All dimensions get the same dilation if only one number is specified.
```
minghuaw, line 50: There's an extra space between "The" and "kernel" in this line.

src/ops/convolution.rs
```diff
@@ -0,0 +1,495 @@
...
+/// - flip: Flip the order in which the spatial dimensions of the weights are processed.
+///   Performs the cross-correlation operator when `flip` is `false` and the convolution
+///   operator otherwise.
+/// - stream: stream or device to evaluate on
+///
+/// # Safety
+///
+/// This function is unsafe because it does not validate the inputs.
+///
+///  # Panic
```
minghuaw, line 106: Does it panic in a way that rust runtime can catch or is it a fatal error that cannot be caught by the rust runtime?

src/ops/convolution.rs
```diff
@@ -0,0 +1,495 @@
...
+
+    let spatial_dims = spatial_dims.unwrap();
+
+    if spatial_dims < 1 || spatial_dims > 2 {
+        return Err(OperationError::WrongInput(
+            "Only 1d and 2d arrays are supported.".to_string(),
+        ));
+    }
+
+    if !array.dtype().is_floating() {
```
minghuaw, line 203: `is_floating` seems weird. I think `is_float` or `is_floating_point` would be better.

src/ops/convolution.rs
```diff
@@ -0,0 +1,495 @@
...
+        ));
+    }
+
+    if weight.ndim() != spatial_dims + 2 {
+        return Err(OperationError::WrongInput(
+            format!("Invalid weight array with {} dimensions for {}d convolution. Expected array with {} dimensions", weight.ndim(), spatial_dims, spatial_dims + 2),
+        ));
+    }
+
+    if array.shape()[spatial_dims + 1] != weight.shape()[spatial_dims + 1] {
```
minghuaw, line 221: Indexing out of bound could panic. `get()` is probably better, it returns `None` if the index is out of bound.

src/ops/convolution.rs
dcvz: Updated to denote what i meant, no #panic header

src/ops/convolution.rs
dcvz: I think this is safely covered by the previous checks but I've used `.get()` anyway.

minghuaw: Maybe we could leave a note that this is already checked and won't panic in the future

PR #52, opened 2024-04-27
 feat(api): indexing + shapes

minghuaw:
Depend on #62


src/ops/shapes.rs
```diff
@@ -0,0 +1,2528 @@
...
+    axis: i32,
+    stream: StreamOrDevice,
+) -> Array {
+    let arrays = arrays.into_iter().collect::<Vec<_>>();
+    stack_inner(&arrays, axis, stream)
+}
+
+#[default_device]
+pub fn try_stack_device(
+    arrays: impl IntoIterator<Item = impl AsRef<Array>>,
```
dcvz, line 1693: Why does this have to be `IntoIterator`? Seems like ` &[impl AsRef<Array>],` would work just as well

minghuaw: True. Not sure why I didn't see this lol

src/ops/shapes.rs
```diff
@@ -0,0 +1,2528 @@
...
+///
+/// ```rust
+/// use mlx::{prelude::*, ops::*};
+///
+/// let x = Array::from_iter(0..4, &[2, 2]);
+/// let y = Array::from_iter(4..8, &[2, 2]);
+/// let result = try_concatenate(&[x, y], 0);
+/// ```
+#[default_device]
+pub fn try_concatenate_device(
```
dcvz, line 570: same as try_stack_device

src/ops/shapes.rs
```diff
@@ -0,0 +1,2528 @@
...
+///
+/// ```rust
+/// use mlx::{prelude::*, ops::*};
+///
+/// let a = Array::from_iter(0..4, &[2, 2]);
+/// let b = Array::from_iter(4..8, &[2, 2]);
+/// let result = try_stack_all(&[&a, &b]);
+/// ```
+#[default_device]
+pub fn try_stack_all_device(
```
dcvz, line 1795: same as try_stack_device

src/utils.rs
```diff
@@ -162,6 +167,45 @@ impl Array {
     }
 }
 
+/// The returned `mlx_vector_array` must be manually freed by the caller.
+pub(crate) unsafe fn new_mlx_vector_array(arrays: &[impl AsRef<Array>]) -> mlx_vector_array {
```
dcvz, line 171: Is it worth wrapping `mlx_vector_array` in an empty-ish Rust struct which implements Drop - just to save ourselves possibly forgetting to free? It seems so far, we use it in a method scope which would all end up dropping the created vector arrays.

src/ops/indexing.rs
```diff
@@ -0,0 +1,1799 @@
...
+    }
+
+    pub(crate) fn start(&self, size: i32) -> i32 {
+        // _start ?? (stride < 0 ? size - 1 : 0)
+
+        // if self.stride.is_negative() {
+        //     self.start + size
+        // } else {
+        //     self.start
+        // }
```
dcvz, line 143: Looks like all these methods have some extra logic comments?

src/ops/indexing.rs
```diff
@@ -0,0 +1,1799 @@
...
+    Ellipsis,
+
+    /// A single index operation
+    ///
+    /// This is equivalent to `arr[1]` in python
+    TakeIndex { index: i32 },
+
+    /// Indexing with an array
+    ///
+    /// An `Rc` is used instead of `Cow` is that even with `Cow`, the compiler will infer
```
dcvz, line 221:
```suggestion
    /// The reason an `Rc` is used instead of `Cow` is that even with `Cow`, the compiler will infer
```

src/array/mod.rs
```diff
@@ -34,6 +35,44 @@ impl std::fmt::Display for Array {
 // TODO: Clone should probably NOT be implemented because the underlying pointer is atomically
 // reference counted but not guarded by a mutex.
 
+impl Clone for Array {
```
minghuaw, line 38: Make this internal use only

src/ops/indexing.rs
```diff
@@ -0,0 +1,1799 @@
...
+
+            Array::from_ptr(c_array)
+        }
+    }
+
+    pub(crate) fn try_slice_device(
+        &self,
+        start: &[i32],
+        stop: &[i32],
+        strides: &[i32],
```
dcvz, line 1045: params here are very similar to `RangeIndex`

src/ops/indexing.rs
```diff
@@ -0,0 +1,1799 @@
...
+                is_slice.push(true);
+
+                let size = shape[i];
+                let absolute_start = range.absolute_start(size);
+                let absolute_end = range.absolute_end(size);
+
+                // // TODO: check if this is correct when stride is negative
+                // let indices: Vec<i32> = (absolute_start..absolute_end)
+                //     .step_by(range.stride().abs() as usize)
+                //     .collect();
```
dcvz, line 1125: is this meant to be resolved or unit tested?

src/ops/indexing.rs
```diff
@@ -0,0 +1,1799 @@
...
+            remaining_indices.extend(operations[(last_array_or_index + 1)..].iter().cloned());
+        }
+    }
+
+    if have_array && remaining_indices.is_empty() {
+        // `clone` returns a new array with the same shape and data
+        match src {
+            Cow::Borrowed(src) => return src.clone(),
+            Cow::Owned(src) => return src,
+        }
```
dcvz, line 1389: no biggie - small suggestion

```suggestion
        return match src {
            Cow::Borrowed(src) => src.clone(),
            Cow::Owned(src) => src,
        }
```

src/ops/shapes.rs
minghuaw: Nvm, there are places where directly accessing the first element (multiple times) is needed

src/ops/indexing.rs
```diff
@@ -0,0 +1,1799 @@
...
+        let mut new_shape = SmallVec::<[i32; 4]>::new();
+        let mut axis_ = 0;
+        for item in remaining_indices {
+            match item {
+                ExpandDims => new_shape.push(1),
+                TakeIndex { .. } => {
+                    if squeeze_needed {
+                        axis_ += 1;
+                    }
+                }
```
dcvz, line 1440: This should be a bug right.. if squeeze is not needed, it should be going into the `_ =>` right?
Suggestion that would make that happen - tests pass in both cases, so I wonder.

```suggestion
                TakeIndex { .. } if squeeze_needed => axis_ += 1,
```

src/ops/indexing.rs
minghuaw: That was already replaced by the absolute_indices line below.

src/utils.rs
dcvz: Let's do it then I'd say!

src/ops/indexing.rs
minghuaw: What do you mean? Your code is doing the same thing

src/ops/indexing.rs
```diff
@@ -0,0 +1,1799 @@
...
+
+    let mut remaining_indices: Vec<ArrayIndexOp> = Vec::new();
+    if have_array {
+        // apply all the operations (except for .newAxis) up to and including the
+        // final .array operation (array operations are implemented via gather)
+        let last_array_or_index = operations
+            .iter()
+            .rposition(|op| op.is_array_or_index())
+            .unwrap(); // safe because we know there is at least one array operation
+                       // let remaining = indices.split_off(last_array_or_index + 1);
```
dcvz, line 1342: we can probably get rid of this right?

src/ops/indexing.rs
dcvz: Not exactly, in my version the if is in  the pattern matching itself, so if `squeeze_needed` is false, it will actually fallthrough into the `_ =>` case. Yours doesn't.

src/ops/indexing.rs
dcvz: it's a commented line:
`// let remaining = indices.split_off(last_array_or_index + 1);`

src/ops/indexing.rs
minghuaw: Right, yours is actually the correct one

src/ops/indexing.rs
minghuaw: > it's a commented line: `// let remaining = indices.split_off(last_array_or_index + 1);`

That line was supposed to go away. Somehow I didn't see that. But the `last_array_or_index` is used later

dcvz: yeah, it was just about the commented line

src/utils.rs
minghuaw: Addressed in 55ca17d931a3dff70b189ecb3086957543ca4d55

src/ops/shapes.rs
minghuaw: Addressed in 55ca17d931a3dff70b189ecb3086957543ca4d55

src/ops/shapes.rs
minghuaw: Addressed in ec661bf335722cc03b7c0da398d6e683b0810bc1

src/ops/shapes.rs
minghuaw: Addressed in ec661bf335722cc03b7c0da398d6e683b0810bc1

src/ops/indexing.rs
minghuaw: Addressed in 55ca17d931a3dff70b189ecb3086957543ca4d55

src/error.rs
```diff
@@ -60,15 +60,166 @@ pub enum FftError {
...
+    #[error("Invalid padding size {size:?} for axis {axis}. Padding sizes must be non-negative")]
+    NegativeWidth { axis: usize, size: (i32, i32) },
+}
+
+#[derive(Debug, Error)]
+pub enum StackError {
+    #[error("No arrays provided for stacking")]
+    NoInputArray,
+
+    #[error(transparent)]
```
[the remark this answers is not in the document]
minghuaw: Addressed in 55ca17d931a3dff70b189ecb3086957543ca4d55

src/ops/indexing.rs
minghuaw: Addressed in 55ca17d931a3dff70b189ecb3086957543ca4d55

src/ops/indexing.rs
minghuaw: Addressed in 55ca17d931a3dff70b189ecb3086957543ca4d55

src/ops/indexing.rs
minghuaw: Addressed in ec661bf335722cc03b7c0da398d6e683b0810bc1

src/ops/indexing.rs
minghuaw: Addressed in ec661bf335722cc03b7c0da398d6e683b0810bc1

src/array/mod.rs
minghuaw: Addressed in 55ca17d931a3dff70b189ecb3086957543ca4d55

src/ops/indexing.rs
```diff
@@ -0,0 +1,1799 @@
...
+
+    unsafe {
+        mlx_sys::mlx_free(indices as *mut _);
+    }
+
+    (max_dims, result)
+}
+
+#[inline]
+fn get_item_int(src: &Array, index: i32, axis: i32, stream: StreamOrDevice) -> Array {
```
dcvz, line 1187: probably should be named: `get_item_index`, right?
Also index, should be getting resolved.

src/ops/indexing.rs
```diff
@@ -0,0 +1,1799 @@
...
+    let ellipsis_count = operations
+        .iter()
+        .filter(|op| matches!(op, ArrayIndexOp::Ellipsis))
+        .count();
+    if ellipsis_count == 0 {
+        return Cow::Borrowed(operations);
+    }
+
+    if ellipsis_count > 1 {
+        panic!("Indexing with multiple ellipsis is not supported");
```
dcvz, line 1228: Should we be bubbling up this error?

src/ops/indexing.rs
minghuaw: > Also index, should be getting resolved.

True. I guess it didn't throw because `mlx` allows out of bound indexing

src/ops/indexing.rs
minghuaw: I don't think so. Indexing isn't supposed to give an `Result` but just panics if indices are invalid

src/ops/indexing.rs
```diff
@@ -0,0 +1,1799 @@
...
+    src: &'a Array,
+    operations: impl Iterator<Item = &'a ArrayIndexOp>,
+    gather_first: bool,
+    stream: StreamOrDevice,
+) -> (usize, Array) {
+    use ArrayIndexOp::*;
+
+    let mut max_dims = 0;
+    let mut slice_count = 0;
+    let mut is_slice: Vec<bool> = Vec::new();
```
dcvz, line 1072: perhaps a small optimization - technically you know the capacity for this

src/ops/indexing.rs
minghuaw: Addressed in 656e701aec731fb757b7ff1a8a20352d8518612f

src/ops/indexing.rs
minghuaw: Not really right? The `operations` iterator is `Filter` which doesn't know the size beforehand

dcvz: Can't you use `size_hint` here?

minghuaw: Well you kind of get the upper bound which would be `last_array_or_index`, but the lower bound would just be 0

minghuaw: Pre-allocates using `last_array_or_index` (7cbeb15bd0f7e98efa10df14e3ba47c18521aa32)

src/ops/indexing.rs
```diff
@@ -0,0 +1,1791 @@
...
+    // `gather_indices` is managed by the `gather_indices` vector.
+    let indices = unsafe {
+        let c_vec_array = mlx_sys::mlx_vector_array_new();
+        for item in gather_indices.iter() {
+            mlx_sys::mlx_vector_array_add(c_vec_array, item.c_array);
+        }
+        c_vec_array
+    };
+    let mut slice_sizes = shape.to_vec();
+    (0..operation_len).for_each(|i| slice_sizes[i] = 1);
```
dcvz, line 1155: I've noticed you've been trying to optimize by reusing the for in operations loop - you can also leverage that for setting these initial values.

src/ops/indexing.rs
```diff
@@ -0,0 +1,1791 @@
...
+            new_shape[i] = item.dim(0);
+            *item = Rc::new(item.reshape(&new_shape));
+        }
+    }
+
+    // Do the gather
+    // let indices = new_mlx_vector_array(gather_indices);
+    // SAFETY: indices will be freed at the end of this function. The lifetime of the items in
+    // `gather_indices` is managed by the `gather_indices` vector.
+    let indices = unsafe {
```
dcvz, line 1147: Could probably make use of `MlxVectorArray` to not have to free later.

dcvz, line 1147: Could probably make use of MlxVectorArray to not have to free later.

src/ops/indexing.rs
dcvz: Do/Should we need to add a unit test here that would've caught this?

src/ops/indexing.rs
minghuaw: Addressed in b7a0ff51933e54218df2811f3cdc1b916cadae5c

src/ops/indexing.rs
minghuaw: Addressed in b7a0ff51933e54218df2811f3cdc1b916cadae5c

src/utils.rs
```diff
@@ -162,6 +169,81 @@ impl Array {
     }
 }
 
+pub(crate) struct MlxVectorArray {
```
dcvz, line 172: One small comment -- we call MlxArray just Array, we can probably call this just VectorArray

src/ops/indexing.rs
minghuaw: > Do/Should we need to add a unit test here that would've caught this?

Maybe put it into the TODO list? Feel free to add one if you can come up with one.

dcvz (approved):
Amazing work here, was no easy task! Let's ship it 🚀

src/utils.rs
minghuaw: Sure. I chose `MlxVectorArray` just so that it kinda tells us it's just a wrapper over c, but I guess `Array` is similar. Addressed in f6632fdaacecd6d521d1308086647f51608abc9a

PR #53, opened 2024-04-29
feat(api): remaining reduction ops

dcvz:


minghuaw (approved):

minghuaw (approved):
LGTM. I think the only thing missing is to document the default value of axes/keepdims if not specified

PR #54, opened 2024-04-30
chore(tooling): enable clippy

dcvz:


.github/workflows/validate.yml
```diff
@@ -10,23 +10,31 @@ concurrency:
...
-#      - name: Run cargo clippy
-#        run: cargo clippy -- -D warnings
+      - name: Run cargo clippy
+        run: cargo clippy -- -D warnings
 
   tests:
     runs-on: blaze/macos-14
     strategy:
       matrix:
         rust: [ stable, 1.75.0 ]
```
minghuaw, line 32: Do we have any reason for this particular version?

minghuaw (approved):
LGTM. Remind me if we've discussed this before, but is there any reason for using 1.75 toolchain in the CI?

dcvz:
> LGTM. Remind me if we've discussed this before, but is there any reason for using 1.75 toolchain in the CI?

We haven't. I based the workflow on a previous where the min supported version was `1.75` due to the main dep having that requirement. we can make it anything we want. point is just that whatever we declare as the min supported version, Ci gets tested for it.

dcvz:
Merged for now, but we can later decide on a minimum version of Rust we want to support

PR #59, opened 2024-05-06
feat(api): eval related updates

dcvz:
- add `try_` and `_unchecked` variants for `::item`
- makes `as_slice` take `&mut` and implicitly call `::eval()`


minghuaw (approved):
LGTM. Didn't realize there were so many `eval`

PR #65, opened 2024-05-20
impl `IndexMutOp`

minghuaw:
### Dependencies

- [ ] #76


dcvz (approved):
tricky one to review, but everything seems to check out and against the other implementations. nice work! 👍

minghuaw:
> tricky one to review, but everything seems to check out and against the other implementations. nice work! 👍

Yeah. This is just like the `Index` PR that it's simply copying the logic and the unit tests from the swift binding

PR #68, opened 2024-05-23
remove stream from param section in docs

minghuaw:
This PR removes the `stream` param from the Params section in the docs.

The same doc is generated for both `_device` and non`_device` variants, and the latter is likely the most commonly used ones. Having `stream` in the docs for the latter would be confusing


dcvz (approved):
🔥 looks good to me!

minghuaw:
> 🔥 looks good to me!

Forgot to update this PR with #67. Will merge this once CI passes after updating this branch

PR #75, opened 2024-06-06
added IntoOption for ergonomics

minghuaw:
A new trait `IntoOption<T>` is added for ergonomics. This trait is just like `Into<Option<T>>` but has special trait impl for conversion from `&[T; N]` into `Option<&[T]>`. This would allow the following use case

```rust
fn foo<'a>(arg: impl IntoOption<&'a [i32]>) { }

fn main() {
    foo(&[1, 2, 3]); // instead of &[1, 2, 3][..]
}
```

Question: Shall we change all occurrence of `&[1, 2, 3][..]` in the test and doc to `&[1, 2, 3]` instead?

### Dependencies

- [x] #76


minghuaw:
This PR is already updated to #76

dcvz (commented):
Love not having [..] - shall we update the usages though?

minghuaw:
> Love not having [..] - shall we update the usages though?

I think most are removed from the doc examples

dcvz (approved):

PR #76, opened 2024-06-08
Update Error Handling

minghuaw:
All c++ exceptions will be caught on the c binding side before crossing the ffi boundary.

This PR changes how exceptions are handled for the ops but functions like `item` and `as_slice` are kept more or unchanged


dcvz (approved):
love it! 🔥

PR #78, opened 2024-06-21
Add remaining arith ops

minghuaw:
TODO:

- [x] port swift binding unit tests for arith ops (https://github.com/ml-explore/mlx-swift/blob/main/Tests/MLXTests/MLXArray%2BOpsTests.swift)


minghuaw:
Tried to add all the remaining arith (and some logical) ops. Might need to double check if we have all the ops

src/ops/logical.rs
```diff
@@ -95,12 +104,16 @@ impl Array {
...
+        &self,
+        other: impl ScalarOrArray<'a>,
+        stream: impl AsRef<Stream>,
+    ) -> Result<Array, Exception> {
         unsafe {
             let c_array = try_catch_c_ptr_expr! {
                 mlx_sys::mlx_greater_equal(
                     self.c_array,
-                    other.c_array,
+                    other.into_owned_or_ref_array().as_ref().as_ptr(),
```
dcvz, line 116: unwieldy! I guess its internal so doesn't matter much but I wonder if there's a better way

src/utils.rs
```diff
@@ -167,3 +168,69 @@ impl<'a, T, const N: usize> IntoOption<&'a [T]> for &'a [T; N] {
...
+
+    fn into_owned_or_ref_array(self) -> Array {
+        self
+    }
+}
+
+impl<'a> ScalarOrArray<'a> for &'a Array {
+    type Array = &'a Array;
+
+    // TODO: clippy would complain about `as_array`. Is there a better name?
```
dcvz, line 189: `as_mlx_array`?

src/utils.rs
```diff
@@ -167,3 +168,69 @@ impl<'a, T, const N: usize> IntoOption<&'a [T]> for &'a [T; N] {
...
+    }
+}
+
+impl ScalarOrArray<'static> for f32 {
+    type Array = Array;
+
+    fn into_owned_or_ref_array(self) -> Array {
+        Array::from_float(self)
+    }
+}
```
dcvz, line 217: can't you do something like what i've done in `random`? 
```
impl<T: ArrayElement> From<T> for ScalarOrArray {
```

dcvz (approved):
LGTM I think! If we haven't gotten them all we'll find out with usage. A few minor questions left on the PR

src/utils.rs
minghuaw: clippy actually complains about the use of `as_` when the argument is `self` instead of `&self`

src/utils.rs
minghuaw: There's a tricky case. A scalar array is treated differently from a singleton matrix (one-element array). One example (but not exhaustive) that I'm aware of right now is `inner` product can broadcast if one arg is a scalar but fails if it's a singleton matrix.

For the scalars, mlx-c only provides bindings for `i32`, `bool`, and `f32`. So it's not as simple as a plain `Into`

minghuaw: For the case with `inner`, the same behavior is observed in numpy, so I don't think that is a bug

src/ops/logical.rs
minghuaw: I guess so

PR #80, opened 2024-06-23
Add remaining index ops

minghuaw:


src/ops/indexing/mod.rs
```diff
@@ -401,36 +401,37 @@ impl Array {
     /// # Params
     ///
     /// - `indices`: The indices to take from the array.
-    /// - `axis`: The axis along which to take the elements. If `None`, the array is treated as a
-    /// flattened 1-D vector.
+    /// - `axis`: The axis along which to take the elements.
```
dcvz, line 404: Did we gain this functionality elsewhere? Seems like we should still be allowing this

minghuaw: It's moved to `take_all`, forgot to check the docs

minghuaw: Resolved in [1f1d65e](https://github.com/oxideai/mlx-rs/pull/80/commits/1f1d65e4d4e147e55517f16587c763e62a10093a)

dcvz (approved):

PR #96, opened 2024-07-16
reorganize workspace

minghuaw:
Reorganizing the workspace for a cleaner structure. The future `mlx-nn` crate and the ergonomics-focused crate would likely depend on the `mlx-rs` crate itself, having a "flat" structure would probably help organizing the dependencies better


minghuaw:
We probably need to change the GH action for building the docs too

mlx-rs/Cargo.toml
```diff
@@ -0,0 +1,46 @@
+[package]
+name = "mlx-rs"
+version = "0.14.0"
+authors = [
+  "Minghua Wu <michael.wu1107@gmail.com>",
+  "David Chavez <david@dcvz.io>",
+]
+edition = "2021"
```
dcvz, line 8: not a big deal but these can be grabbed from workspace Cargo.toml

dcvz (approved):

mlx-rs/Cargo.toml
minghuaw: Didn't see this before I hit the squash and merge button lol. We could do this later then.

PR #100, opened 2024-08-02
Initial attempt at implementing `mlx-nn` NN modules

minghuaw:
This is my initial attempt at implementing the neural net modules and optimizers. The following are included in this PR

- `nn/activation`
- `nn/convolution`
- `nn/dropout`
- `nn/linear`
- `nn/sequential`
- `nn/losses/`
- `optimizer/sgd`
- `optimizer/rmsprop`


minghuaw:
This is just an initial attempt at implementing some of the neural net components. I think some feedback on the overall API design and ergonomics would be nice

minghuaw:
Ended up changing `Module` trait to take `&Array`. The only cost (so far) is that an empty`Sequential` would end up `deep_clone` the input array, but this should make overall usage more flexible as most (if not all) ops only require a ref input

minghuaw:
Still missing a bunch of docs and tests, but I think we could start reviewing some key impls.

1. `mlx_nn::value_and_grad::value_and_grad()` function. This is where I had the discussion about whether `model.update(...)` in the swift binding is necessary, and it turns out that we need this too. Essentially, if we don't update the model's parameters to the ones passed in, the gradients will be all zeros. This, however, means the closure returned by `value_and_grad` requires a mutable reference to the model, which seems a bit counter intuitive (why would computing the value and gradients require a mutable borrow?). So I guess the question is whether there's a better way to handle this or should we just clarify in the docs?
2. Traits and macros related to `Module`. The `Module` trait is separated into a new crate `mlx-nn-module` so that it would be easier to handle the dep problem in the the derive macro impl. We need to use this macro in `mlx-nn` but the users should be able to use this macro in their own crates as well.

minghuaw:
@dcvz It would be nice if #108 can be reviewed first. I'd like to see if some of the SIGSEGV was only caused by the `ScalarOrArray`.

mlx-nn/src/losses.rs
```diff
@@ -0,0 +1,1281 @@
...
+            LossReduction::Sum => Ok(loss.sum(None, None)?),
+            LossReduction::Mean => Ok(loss.mean(None, None)?),
+        }
+    }
+}
+
+/// Optional parameters for the `cross_entropy` function.
+#[derive(Debug, Clone, Default)]
+#[option_builder]
+pub struct CrossEntropyOptions<'a> {
```
minghuaw, line 58: Tried a different way to handle optional args. Took this approach recommended in azure sdk guideline for rust. Not too sure if this is an overkill.

@dcvz WDYT

dcvz: It's interesting for sure.. it's not too bad -- remind's a bit of Burn's builders. Have you seen those? The syntax there IMO is slightly better, but I think the ideas are very similar.

Here's for example a declaration: https://github.com/oxideai/diffusers-burn/blob/d52deb45110f721d8e8cb347d0e56ef80c56fc28/src/models/attention.rs#L94

and the usage here: https://github.com/oxideai/diffusers-burn/blob/d52deb45110f721d8e8cb347d0e56ef80c56fc28/src/models/attention.rs#L257

WDYT?

examples/mnist/src/mnist.rs
```diff
@@ -0,0 +1 @@
+// TODO
```
dcvz, line 1: :D

mlx-macros/src/lib.rs
```diff
@@ -142,3 +145,63 @@ pub fn generate_test_cases(input: TokenStream) -> TokenStream {
 
     TokenStream::from(tests)
 }
+
+/// Derive the `ModuleParameters` trait for a struct. Mark a field with `#[param]` attribute to
```
dcvz, line 149: Do you think we'll eventually want to have a `derive(Module)`?

mlx-nn/src/optimizers/rmsprop.rs
```diff
@@ -0,0 +1,141 @@
...
+            state: OptimizerState::new(),
+        }
+    }
+
+    /// Sets the smoothing constant. Default to [`RmsProp::DEFAULT_ALPHA`] if not specified.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `alpha` is negative.
+    pub fn with_alpha(mut self, alpha: impl Into<Option<f32>>) -> Self {
```
dcvz, line 55: Should we type this to not be negative? I'm always a bit torn on that kind of typing

mlx-nn/src/optimizers/rmsprop.rs
```diff
@@ -0,0 +1,141 @@
...
+        assert!(epsilon >= 0.0);
+        self.epsilon = epsilon;
+        self
+    }
+}
+
+impl Optimizer for RmsProp {
+    fn update_single(&mut self, key: Rc<str>, gradient: Array, parameter: &mut Array) {
+        // let v = alpha * state + (1 - alpha) * square(gradient)
+        // return (parameter - learningRate * gradient / (sqrt(v) + eps), v)
```
dcvz, line 79: We can get rid of this right?

mlx-nn/src/utils.rs
```diff
@@ -0,0 +1,118 @@
...
+
+use std::rc::Rc;
+
+use mlx_rs::module::FlattenedModuleParam;
+use mlx_rs::Array;
+
+/// A custom type to indicate whether a `Module` should include a bias or not.
+/// Default to `Yes`.
+#[derive(Debug, Clone, Copy, Default)]
+pub enum WithBias {
```
dcvz, line 11: Will be nice to eventually be able to define bias as `Tensor<Shape>`.. I wonder if there's some way we might already be able to do something similar with Array

mlx-rs/src/transforms/mod.rs
```diff
@@ -231,11 +248,12 @@ fn build_gradient<'a, F>(
 where
     F: FnMut(&[Array]) -> Vec<Array> + 'a,
 {
+    let argument_numbers = argument_numbers.into_option().unwrap_or(&[0]);
     let closure = Closure::new(f);
     build_gradient_inner(closure, argument_numbers)
 }
 
-fn build_gradient_fallible<'a, F>(
+fn build_fallible_gradient<'a, F>(
```
dcvz, line 256: def like the faillible as prefix better!

dcvz (commented):
Looks good overall! Let's align on the optionals so we can have one way of doing things but I like the direction for them too.

I have the gut feeling we may have a better way to do some of the utils for types we're using, but i'm happy to try and improve that once we start writing more examples and see where it could be more ergonomic.

mlx-nn/src/losses.rs
minghuaw: Interesting. Their `CrossEntropy` is a struct and the actual computation takes place in `fn forward`. We probably don't want to call it `forward` just to avoid confusion with the `Module` trait if we were to adapt their approach?

mlx-macros/src/lib.rs
minghuaw: I don't think so because the `Module` implementation would include the `forward` method. The associated error type was an attempt to allow user to have more customized errors but is currently limited because the c ffi can only take an mlx `Exception`. One thing I think that we could change to the `Module` trait is to provide an default noop `training_mode` implementation, but I didn't do this because I felt like user might forget if this is not mandatory.

mlx-nn/src/optimizers/rmsprop.rs
minghuaw: Is there an existing crate that does this for `f32` or should we add a new type?

mlx-nn/src/optimizers/rmsprop.rs
minghuaw: That was me trying to make sure the math is correct, but yeah we could get rid of this

examples/mnist/src/mnist.rs
minghuaw: haha, this mod would be just downloading the mnist dataset and provide a loader

mlx-nn/src/utils.rs
minghuaw: This is a part that I'm not too sure about. This, on the on hand, is not consistent with how we handle optional arg in other `Module`s or functions, but on the other hand the bias would either be `zeros` or `random::uniform`. But we could probably get rid of this and just define an associated `pub const` for the conv layers and linear layers (these are the only `Module`s that use this type right now

mlx-nn/src/convolution.rs
```diff
@@ -0,0 +1,456 @@
...
+    ops::{conv1d, conv2d, zeros},
+    random::uniform,
+    Array,
+};
+
+use crate::utils::{IntOrPair, IntOrTriple, WithBias};
+
+/// Optional parameters for the `Conv1d` module.
+#[derive(Debug, Clone, Default)]
+pub struct Conv1dBuilder {
```
minghuaw, line 14: @dcvz What about this kind of builder pattern? We could get rid of the `WithBias` type this way, and we could probably apply this on all `Module`'s that take optional args? For `cross_entropy`, we could make it a struct and then apply this approach?

minghuaw: Or we could have two different way of handling optional args? Use this builder pattern on all `Module` structs, and keep functions like `cross_entropy` as it is?

mlx-nn/src/activation.rs
```diff
@@ -265,11 +265,11 @@ pub fn hard_swish(x: impl AsRef<Array>) -> Result<Array, Exception> {
 ///
 /// This splits the `axis` dimension of the input into two halves
 /// (`a` and `b`) and applies `a * sigmoid(b)`.
-#[derive(Debug, Clone, Default, ModuleParameters)]
-#[option_builder]
+#[derive(Debug, Clone, Default, ModuleParameters, GenerateBuilder)]
 pub struct Glu {
     /// The axis to split the input tensor. Default to [`Glu::DEFAULT_AXIS`] if not provided.
-    pub axis: Option<i32>,
+    #[optional(default_value = Glu::DEFAULT_AXIS)]
```
dcvz, line 271: do we support `default_value` for non optionals?  Otherwise it might be simpler for us to just keep the `Optional<T>` type and just annotate with `#[default_value()]`

mlx-macros/tests/test_generate_builder.rs
```diff
@@ -0,0 +1,29 @@
+use mlx_macros::GenerateBuilder;
+
+#[derive(Debug, GenerateBuilder)]
+struct TestStruct {
+    #[optional(default_value = TestStruct::DEFAULT_OPT_FIELD_1)]
```
dcvz, line 5: do we support default_value for non optionals? Otherwise it might be simpler for us to just keep the Optional<T> type and just annotate with #[default_value()]

mlx-macros/tests/test_generate_builder.rs
```diff
@@ -0,0 +1,29 @@
...
+    opt_field_1: i32,
+    #[optional(default_value = TestStruct::DEFAULT_OPT_FIELD_2)]
```
dcvz, line 7: Do you think it's still useful to have consts for defaults here? The fact that we can annotate means we can quickly see what those defaults are going to be without the indirection. Or are you worried about perf?

mlx-nn/src/activation.rs
```diff
@@ -0,0 +1,1554 @@
...
+impl Glu {
+    /// The default axis value.
+    pub const DEFAULT_AXIS: i32 = -1;
+}
+
+impl Module for Glu {
+    type Error = Exception;
+
+    fn forward(&self, x: &Array) -> Result<Array, Self::Error> {
+        println!("axis: {}", self.axis);
```
dcvz, line 290: stray print?

mlx-macros/tests/test_generate_builder.rs
dcvz: Also because this is a macro, IDEs won't always be able to CMD+click into the definition.

mlx-macros/src/lib.rs
```diff
@@ -142,3 +145,63 @@ pub fn generate_test_cases(input: TokenStream) -> TokenStream {
...
+        const _: () = {
+            extern crate mlx_rs as _mlx_rs;
+            #module_param_impl
+        };
+    };
+    TokenStream::from(output)
+}
+
+#[proc_macro_derive(GenerateBuilder, attributes(generate_builder, optional))]
+pub fn derive_generate_builder(input: TokenStream) -> TokenStream {
```
dcvz, line 203: It seems like very often users of `GenerateBuilder` also derive Default, do you think that's usually the case?

PR #102, opened 2024-09-04
Trust synchronization on the cpp side

minghuaw:
This PR removes all methods that require mutable borrows on `Array`. We might have been too cautious, but it seems like all the computation takes place when `eval` is called, and the `eval` event on the cpp side is protected by mutex and cond var (https://github.com/ml-explore/mlx/blob/969337345ffd34159884c8bf124dbb6eeaa902bb/mlx/backend/no_metal/event.cpp#L21)


dcvz (approved):
LGTM 👍

PR #115, opened 2024-09-25
Experimental mlx c fallible closure

minghuaw:
This is simply to show how the fallible closure problem could be resolved if the PR to the c binding (https://github.com/ml-explore/mlx-c/pull/35) were accepted

PR #116, opened 2024-09-30
Use custom c shim for fallible closure

minghuaw:
This is probably a better approach than #115 as it doesn't require modification to the upstream c binding.


dcvz (approved):
LGTM 👍

PR #120, opened 2024-10-22
Change `GenarateBuilder` and `GenerateTestCases` to function-like proc_macro

minghuaw:
I think a derive macro generally implies implementation of a trait of the same name, which is not true in the case of `GenerateBuilder` and `GenerateTestCases`. This PR changes them to function-like proc_macro to avoid such confusion. Functional changes include:

1. The `generate_builder!()` macro now checks if the struct it's applied to derives Default and triggers a panic if it does.


dcvz (approved):
LGTM 👍

PR #121, opened 2024-10-26
feat(nn): embedding

dcvz:


mlx-nn/src/embedding.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub fn as_linear(&self, x: &Array) -> Result<Array, Exception> {
+        mlx_rs::ops::matmul(x, &self.weight.value.t())
+    }
+}
+
+impl Module for Embedding {
+    type Error = Exception;
+
+    fn forward(&self, x: &Array) -> Result<Array, Self::Error> {
+        Ok(self.weight.index(x.clone()))
```
minghuaw, line 51: Should we use `x` as index here as using `&Array` as index is already supported

mlx-nn/src/embedding.rs
```diff
@@ -0,0 +1,70 @@
...
+    fn training_mode(&mut self, _mode: bool) {}
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use mlx_rs::array;
+
+    #[test]
+    fn test_embedding() {
```
minghuaw, line 63: Can we port the unit test from the swift binding (https://github.com/ml-explore/mlx-swift/blob/d649c62b77c487c25012910b0d02b30283d388ca/Tests/MLXTests/IntegrationTests.swift#L6843)

mlx-nn/src/lib.rs
```diff
@@ -15,6 +15,7 @@ mod activation;
 mod container;
 mod convolution;
 mod dropout;
+pub mod embedding;
```
minghuaw, line 18: Maybe we should be consistent that all layers are re-exported to the crate root instead of their own separate mod?

minghuaw (commented):
Looks good overall. Left a few comments

minghuaw (approved):
Let's merge it!

PR #122, opened 2024-10-26
Allow indexing with `&Array`

minghuaw:
This PR adds support for using `&Array` as index. Other changes include removing the marker trait `IndexBounds` and replaced with impls with explicit types instead of `T: IndexBounds`

### DIFF
--- a/mlx-rs/src/ops/indexing/mod.rs
+++ b/mlx-rs/src/ops/indexing/mod.rs
@@ -12,7 +12,7 @@
 //! |------|-------------|
 //! | `i32` | An integer index |
 //! | `Array` | Use an array to index another array |
-//! | `Rc<Array>` | Use an array to index another array |
+//! | `&Array` | Use a reference to an array to index another array |
 //! | `std::ops::Range<i32>` | A range index |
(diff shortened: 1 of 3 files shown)


mlx-rs/src/ops/indexing/mod.rs
```diff
@@ -283,25 +293,25 @@ pub trait IndexMutOp<Idx, Val> {
     }
 }
 
-/// A marker trait for range bounds that are `i32`.
-pub trait IndexBounds: RangeBounds<i32> {}
+// /// A marker trait for range bounds that are `i32`.
```
dcvz, line 296: Shall we remove the commented out ones?

dcvz (approved):
LGTM just a small comment

mlx-rs/src/ops/indexing/mod.rs
minghuaw: Yeah, forgot to remove those lines. Resolved in [bfe8071](https://github.com/oxideai/mlx-rs/pull/122/commits/bfe80719e6306a96493372525ab49d6065d8cd2e)

PR #126, opened 2024-10-30
chore(mlx-c): Update MLX-C to 0.10.0

dcvz:
Updates MLX-C to 0.0.10:
- fixes shims
- adds new types for changed methods
- adds padding mode for pad methods

It does not add new additions, those will be tackled separately.

### DIFF
--- a/mlx-rs/src/ops/shapes.rs
+++ b/mlx-rs/src/ops/shapes.rs
@@ -1183,10 +1183,10 @@ mod tests {
     #[test]
     fn test_pad() {
         let x = Array::zeros::<f32>(&[1, 2, 3]).unwrap();
-        assert_eq!(pad(&x, 1, None).unwrap().shape(), &[3, 4, 5]);
-        assert_eq!(pad(&x, (0, 1), None).unwrap().shape(), &[2, 3, 4]);
+        assert_eq!(pad(&x, 1, None, None).unwrap().shape(), &[3, 4, 5]);
+        assert_eq!(pad(&x, (0, 1), None, None).unwrap().shape(), &[2, 3, 4]);
         assert_eq!(
-            pad(&x, &[(1, 1), (1, 2), (3, 1)], None).unwrap().shape(),
+            pad(&x, &[(1, 1), (1, 2), (3, 1)], None, None).unwrap().shape(),
             &[3, 5, 7]
         );
     }
--- a/mlx-rs/src/ops/shapes.rs
+++ b/mlx-rs/src/ops/shapes.rs
@@ -660,6 +660,24 @@ impl<'a> PadWidth<'a> {
     }
 }
 
+#[derive(Debug)]
+pub enum PadMode {
+    Constant,
+    Edge,
+}
+
+impl PadMode {
+    fn as_mlx_string(&self) -> mlx_sys::mlx_string {
+        static CONSTANT: &[u8] = b"constant\0";
+        static EDGE: &[u8] = b"edge\0";
+
+        match self {
+            PadMode::Constant => unsafe { mlx_sys::mlx_string_new(CONSTANT.as_ptr() as *const _) },
+            PadMode::Edge => unsafe { mlx_sys::mlx_string_new(EDGE.as_ptr() as *const _) },
+        }
+    }
+}
+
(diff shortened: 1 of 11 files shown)


mlx-rs/src/ops/shapes.rs
```diff
@@ -677,13 +695,14 @@ impl<'a> PadWidth<'a> {
 /// use mlx_rs::{prelude::*, ops::*};
 ///
 /// let a = Array::from_iter(0..4, &[2, 2]);
-/// let result = pad(&a, 1, Array::from_int(0));
+/// let result = pad(&a, 1, Array::from_int(0), PadMode::Constant);
```
minghuaw, line 698: Is this the only place where `PadMode` is used? If so, should we just make `PadMode::Constant` the default value?

mlx-rs/src/ops/other.rs
```diff
@@ -91,13 +92,50 @@ pub fn diagonal_device(
...
+/// - subscripts: Einstein summation convention equation
+/// - operands: input arrays
+/// - stream: stream or device to evaluate on
+#[default_device]
+pub fn einsum_device(
+    subscripts: &str,
+    operands: &[&Array],
+    stream: impl AsRef<Stream>,
+) -> Result<Array, Exception> {
+    let subscripts = unsafe { mlx_sys::mlx_string_new(subscripts.as_ptr() as *const i8) };
```
minghuaw, line 108: Do we need to check if the string is null terminated here?

mlx-rs/src/ops/other.rs
```diff
@@ -91,13 +92,50 @@ pub fn diagonal_device(
...
+///
+/// # Params
+///
+/// - subscripts: Einstein summation convention equation
+/// - operands: input arrays
+/// - stream: stream or device to evaluate on
+#[default_device]
+pub fn einsum_device(
+    subscripts: &str,
+    operands: &[&Array],
```
minghuaw, line 105: We could probably make this something like `impl IntoIterator<Item = &Array>` since all it needs right now is iter and collect

mlx-rs/src/ops/convolution.rs
```diff
@@ -191,6 +191,149 @@ pub fn conv3d_device(
     }
 }
 
+/// 1D transposed convolution over an input with several channels.
+///
+/// > Only the default `groups=1` is currently supported.
```
minghuaw, line 196: No biggy, just wondering how `>` is rendered in the generated docs

mlx-rs/src/ops/convolution.rs
```diff
@@ -191,6 +191,149 @@ pub fn conv3d_device(
...
+/// - array: input array of shape `[N, H, C_in]`
+/// - weight: weight array of shape `[C_out, H, C_in]`
+/// - stride: kernel stride
+/// - padding: input padding
+/// - dilation: kernel dilation
+/// - groups: input feature groups
+/// - stream: stream or device to evaluate on
+#[default_device]
+pub fn conv_transposed1d_device(
+    array: &Array,
```
minghuaw, line 209: Shall we make all `&Array` just `impl AsRef<Array>` to be more ergonomic-ish?

mlx-rs/src/ops/convolution.rs
```diff
@@ -191,6 +191,149 @@ pub fn conv3d_device(
...
+    weight: &Array,
+    stride: impl Into<Option<i32>>,
+    padding: impl Into<Option<i32>>,
+    dilation: impl Into<Option<i32>>,
+    groups: impl Into<Option<i32>>,
+    stream: impl AsRef<Stream>,
+) -> Result<Array, Exception> {
+    let stride = stride.into().unwrap_or(1);
```
minghuaw, line 217: We should probably mention the default values in the doc (should probably do the same for `conv_transpose2d` and `conv_transpose3d`

minghuaw (commented):
Good work! Didn't realize they have this many changes in mlx-c. Left some comments, but looks good overall. I'm thinking given that since this does not include the additional ops that were introduced upstream, maybe we can still merge it if the clippy is only complaining about things like `dead_code` and `unused_variables`

PR #128, opened 2024-10-31
chore(mlx-c): Add remaining additions

dcvz:
This is the remaining additions to the core.

### DIFF
--- a/mlx-rs/src/ops/indexing/mod.rs
+++ b/mlx-rs/src/ops/indexing/mod.rs
@@ -320,17 +320,32 @@ impl Array {
     pub fn take_device(
         &self,
         indices: &Array,
-        axis: impl Into<Option<i32>>,
+        axis: i32,
         stream: impl AsRef<Stream>,
     ) -> Result<Array, Exception> {
         unsafe {
-            let c_array = match axis.into() {
-                Some(axis) => try_catch_c_ptr_expr! {
-                    mlx_sys::mlx_take(self.c_array, indices.c_array, axis, stream.as_ref().as_ptr())
-                },
-                None => try_catch_c_ptr_expr! {
-                    mlx_sys::mlx_take_all(self.c_array, indices.c_array, stream.as_ref().as_ptr())
-                },
+            let c_array = try_catch_c_ptr_expr! {
+                mlx_sys::mlx_take(self.c_array, indices.c_array, axis, stream.as_ref().as_ptr())
+            };
+
+            Ok(Array::from_ptr(c_array))
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


mlx-rs/src/ops/indexing/mod.rs
```diff
@@ -320,65 +320,88 @@ impl Array {
...
         stream: impl AsRef<Stream>,
     ) -> Result<Array, Exception> {
         unsafe {
-            let c_array = try_catch_c_ptr_expr! {
-                mlx_sys::mlx_take(self.c_array, indices.c_array, axis, stream.as_ref().as_ptr())
+            let c_array = match axis.into() {
+                Some(axis) => try_catch_c_ptr_expr! {
+                    mlx_sys::mlx_take(self.c_array, indices.c_array, axis, stream.as_ref().as_ptr())
+                },
+                None => try_catch_c_ptr_expr! {
```
minghuaw, line 331: Don't we already have `take_all`? https://github.com/oxideai/mlx-rs/blob/aac7e03e609b20b5b28aae3bb5eef61dcc2319f9/mlx-rs/src/ops/indexing/mod.rs#L341 

I think it was intentional that we separated `take` and `take_all`

minghuaw: I think we separated `take` and `take_all` because this pattern also appears in other functions, for example `argsort` vs `argsort_all`, `stack` vs `stack_all`. I'm thinking we probably want to be consistent on this? Either all these functions don't have a `*_all` variants or all of them do?

mlx-rs/src/ops/shapes.rs
```diff
@@ -688,6 +688,7 @@ impl PadMode {
 ///   `(before_i, after_i)` are all the same. If a single integer or tuple with a single integer is
 ///   passed then all axes are extended by the same number on each side.
 /// - `value`: The value to pad the array with. Default is `0` if not provided.
+/// - `mode`: The padding mode. Default is `PadMode::Constant` if not provided.
```
minghuaw, line 691: Should `PadMode` just implement the `Default` trait too?

minghuaw (commented):
I think the only concern is whether we should merge `take` and `take_all` into one function. The rest looks good to me!

I remember there was something about distributed ops at some point. Are those still something we need to include?

mlx-rs/src/ops/shapes.rs
minghuaw: true, let's just leave it as is then

minghuaw (approved):
LGTM!

PR #131, opened 2024-10-31
feat(mlx-fast): Implement NN part of mlx-fast

dcvz:
Implements the speedup NN methods - does not yet implement custom kernels.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,6 @@ mlx-macros = { version = "0.1.0", path = "mlx-macros" }
 mlx-internal-macros = { version = "0.1.0", path = "mlx-internal-macros" }
 mlx-rs = { version = "0.14.0", path = "mlx-rs" }
 mlx-nn = { version = "0.14.0", path = "mlx-nn" }
-mlx-fast = { version = "0.14.0", path = "mlx-fast" }
 
 # external dependencies
 thiserror = "1"


mlx-fast/src/lib.rs
```diff
@@ -0,0 +1,267 @@
+use mlx_internal_macros::default_device;
+use mlx_rs::{Array, Stream, StreamOrDevice};
+
+// TODO: Implement custom kernels:
+// https://github.com/ml-explore/mlx-swift/blob/d649c62b77c487c25012910b0d02b30283d388ca/Source/MLXFast/MLXFastKernel.swift
+
+/// Optimized implementation of `NN.RoPE`.
+#[allow(non_snake_case)]
+#[default_device]
+pub fn RoPE_device<'a>(
```
minghuaw, line 11: Function name should be snake case right? So `rope_device`?

mlx-fast/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+    stream: impl AsRef<Stream>,
+) -> Array {
+    let base = base.into();
+    let base = mlx_sys::mlx_optional_float {
+        value: base.unwrap_or(0.0),
+        has_value: base.is_some(),
+    };
+
+    unsafe {
+        let new_array = mlx_sys::mlx_fast_rope(
```
minghuaw, line 28: I haven't looked into the c++ side. Just curious, do we need to catch errors here?

mlx-fast/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+///
+/// Supports [Multi-Head Attention](https://arxiv.org/abs/1706.03762), [Grouped Query Attention](https://arxiv.org/abs/2305.13245), and [Multi-Query Attention](https://arxiv.org/abs/1911.02150).
+///
+/// This function will dispatch to an optimized Metal kernel when the query sequence length is 1. It handles other cases with regular MLX operations.
+///
+/// > Note: The softmax operation is performed in float32 precision regardless of input precision (float16 or float32).
+///
+/// > Note: For Grouped Query Attention and Multi-Query Attention, the input arrays for `key` and `value` should not be pre-tiled to match the `query` array.
+#[default_device]
+pub fn scaled_dot_product_attention_device<'a>(
```
minghuaw, line 56: The result is not returned

mlx-fast/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+/// - stream: stream or device to evaluate on
+#[default_device]
+pub fn rms_norm_device<'a>(
+    x: impl AsRef<Array>,
+    weight: impl AsRef<Array>,
+    eps: f32,
+    stream: impl AsRef<Stream>,
+) -> Array {
+    unsafe {
+        let new_array = mlx_sys::mlx_fast_rms_norm(
```
minghuaw, line 104: We should catch error here right? The c++ side would throw if the shapes don't match wouldn't it?

mlx-fast/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+/// - eps: A small additive constant for numerical stability
+/// - stream: stream or device to evaluate on
+#[default_device]
+pub fn layer_norm_device<'a>(
+    x: impl AsRef<Array>,
+    weight: impl Into<Option<&'a Array>>,
+    bias: impl Into<Option<&'a Array>>,
+    eps: f32,
+    stream: impl AsRef<Stream>,
+) -> Array {
```
minghuaw, line 135: Similarly, should we catch error here?

mlx-fast/src/lib.rs
minghuaw: The same applies to other functions

minghuaw (approved):
LGTM!

PR #136, opened 2024-11-04
feat(nn): Add positional encoding

dcvz:


mlx-nn/src/positional_encoding.rs
```diff
@@ -0,0 +1,96 @@
...
+    ///
+    /// The traditional implementation rotates consecutive pairs of elements in the
+    /// feature dimension while the default implementation rotates pairs with
+    /// stride half the feature dimensions for efficiency.
+    ///
+    /// For more details see _RoFormer: Enhanced Transformer with Rotary Position
+    /// Embedding_ ([https://arxiv.org/abs/2104.09864](https://arxiv.org/abs/2104.09864))
+    #[derive(Debug, Clone, ModuleParameters)]
+    #[generate_builder(generate_build_fn = true)]
+    struct RoPE {
```
minghuaw, line 18: I feel like `Rope` probably fits the rust naming convention better?

minghuaw: Or maybe we can do `pub struct RotaryPositionEncoding` and then add an alias `pub type Rope = RotaryPositionalEncoding`?

minghuaw:
Closing in favour of #172

PR #139, opened 2024-11-12
feat(nn): Add pooling

minghuaw:


mlx-nn/src/pooling.rs
```diff
@@ -0,0 +1,338 @@
...
+    /// Stride of the pooling window
+    stride: Vec<usize>,
+
+    /// Axes to pool over
+    axes: Vec<i32>,
+
+    /// Pooling operation
+    ///
+    /// TODO: We have Arc here just to make it `Clone` and `Send`. Is this necessary?
+    pooling_op: Arc<PoolingOp>,
```
minghuaw, line 36: One thing I'm not sure here is whether we should use `Arc`. My thinking right now is that this is going to be a trait object anyway and might as well just make it `Clone` and `Send`

PR #140, opened 2024-11-12
feat(nn): Add quantization

minghuaw:


minghuaw:
Splitting into two separate PRs

PR #142, opened 2024-11-16
feat(nn): Part 1 of quantization

minghuaw:
This PR depends on #143 to be merged first. The `QuantizedLinear` is to be created with all params frozen which would have a nicer impl if #143 is merged


minghuaw:
> Regarding what to test @minghuaw : https://github.com/ml-explore/mlx/blob/main/python/tests/test_quantized.py Does that work?

Should work. Will port those tests in a new PR

PR #148, opened 2024-11-20
exploration: builder macro

dcvz:


minghuaw:
I like the overall direction. 👍 

A few general comments.

1. I think we should avoid using derive macro if it's not actually implementing a new trait.
2. Shall we integrate this with the current `generate_builder` macro?
3. I would argue the the macro should add the `Default` impl by just calling `<Builder>::new()`. I think clippy will actually complain if you have a `new()` that doesn't take any args but not implementing `Default`
4. Is this supposed to be a macro for general public users or is it intended to be used internally only?

minghuaw:
Integrated in #150

PR #156, opened 2024-12-02
chore: Fix latest clippy warnings

minghuaw:
Seems there's a lot new lints added upstream. This PR simply fixes the new warnings


minghuaw:
This is not needed once we merge #147

PR #157, opened 2024-12-14
Exploration: use macro to mimic named args for optional args

minghuaw:
Example usage

```rust
#![allow(unused_variables)]

use mlx_internal_macros::{default_device, generate_macro};
use mlx_rs::{Stream, StreamOrDevice};

/// Test macro generation.
#[generate_macro]
#[default_device]
fn foo_device(
    a: i32,                                 // Mandatory argument
    b: i32,                                 // Mandatory argument
    #[optional] c: Option<i32>,             // Optional argument
    #[optional] d: impl Into<Option<i32>>,  // Optional argument but impl Trait
    #[optional] stream: impl AsRef<Stream>, // stream always optional and placed at the end
) -> i32 {
    a + b + c.unwrap_or(0) + d.into().unwrap_or(0)
}

#[test]
fn test_foo() {
    assert_eq!(foo!(1, 2), 3);
    assert_eq!(foo!(1, 2, c = Some(3)), 6);
    assert_eq!(foo!(1, 2, d = Some(4)), 7);
    assert_eq!(foo!(1, 2, c = Some(3), d = Some(4)), 10);

    let stream = Stream::new();

    assert_eq!(foo!(1, 2, stream = &stream), 3);
    assert_eq!(foo!(1, 2, c = Some(3), stream = &stream), 6);
    assert_eq!(foo!(1, 2, d = Some(4), stream = &stream), 7);
    assert_eq!(foo!(1, 2, c = Some(3), d = Some(4), stream = &stream), 10);
}
```


minghuaw:
> Haven't looked deeply, but looks cool from the API level!

Any thoughts on the ergonomics side? Like should the mandatory args also be named or just left as positional args?

minghuaw:
@dcvz Mind taking a deeper look at this (maybe before 0.22)?

minghuaw:
> 1. How big is the number of generated methods across all functions and does that negatively affect compile times / editor functionality?

The size of the generated macro would depend on the number of optional args. It runs a permutation of all possible combinations so this could indeed escalate fast. This would definitely slow down the compile time, but it's something I haven't really measured yet. Will try to come up with a way to test this.

> 2. I'm a bit torn about the fact that including this would force us to split ops into a separate package. It's probably not a big deal but it always feels weird to be forced into something by technological choices. Have we ever seen another project do something similar to this?

We probably don't want to split it into separate crate. I've attempted to split the ops from the core part that has `Array` and all other utilities, and I was running into circular dependencies. There might be ways to get around the circular dep problem but splitting is still going to take quite some time given how interleaved the ops and the base utilities are right now.

> More of a side note: is it worth updating one of our examples in this draft to use this new helper syntax? That might also be a nice way to give it application, and test it out in the wild.

We definitely should (and the doc examples as well). I feel like we still need to add a lot of stuff in the docs.

I'm gonna try to use this macro on a few actual functions and see how that turns out

minghuaw:
We might want to change functions like `arange<T, U>` to use a single generic that can be inferred from the args. This would simplify integrating this macro

minghuaw:
Applied the macro on most of the ops functions, and added a new feature to the macro to allow named mandatory field, which is used when a mandatory arg is placed after an optional arg (see the `arange` example below)

Definition

```rust
/// See [`Array::arange`]
#[generate_macro]
#[default_device]
pub fn arange_device<U, T>(
    #[optional] start: impl Into<Option<U>>,
    #[named] stop: U,
    #[optional] step: impl Into<Option<U>>,
    #[optional] stream: impl AsRef<Stream>,
) -> Result<Array>
where
    U: NumCast,
    T: ArrayElement,
{
    Array::arange_device::<U, T>(start, stop, step, stream)
}
```

Usage

```rust
let array = Array::arange::<_, f32>(None, 50, None).unwrap();
let array = mlx_rs::arange!(stop = 50).unwrap();
```

The test `mlx-tests/tests/test_exported_macro.rs` has more examples (but this file right now only tests a few functions from different mods just to ensure the path to the function in the macro is correct

minghuaw:
> I don't love that we have to mark `#[optional]` everywhere now.. and perhaps in the future we can find a better alternative for this. however, i wouldn't mark this as blocking.

It is indeed too repetitive. Need to see if there's a good way to determine the type other than just the ident.

PR #159, opened 2024-12-17
feat(example): update mnist example

minghuaw:
Shall we include the whole dataset? The `mnist` crate can download the dataset but doesn't always work during testing


minghuaw:
> Ideally would be better to download at test time, but I think for now is ok!

Agree. The `mnist` crate actually supports downloading at runtime but it was surprisingly slow when I tested it. Didn't get time to investigate that but we can revisit this in the future

PR #160, opened 2024-12-18
Experiment with caching to improve perf

minghuaw:
Major changes:

1. Make `mlx-nn` a mod in `mlx-rs`
2. Change `Param` type to be a wrapper around `RefCell<T>` and check mut access at runtime


minghuaw:
This experiment was started because we observed performance difference in the rust version and python version of the mnist example. We are closing this because it was found that caching is not needed to fix the perf issue. Below is a summary of the cause of the perf diff:

1. `Sgd` optimizer used logical ops on scalar arrays, which can be 10 to 100 times slower than just creating a new scalar array
2. The naive loading function in the rust version was repeatedly stacking arrays for every iteration instead of collecting the stacked arrays into a vec and then iterate
3. The simple two layer MLP used in the example runs faster on CPU than GPU, and the python example was running on cpu by default while the rust example was running on gpu. When switched to the same default device, no significant perf difference was observed (after the first two items was fixed)

PR #171, opened 2025-01-05
feat(nn): Another attempt at positional encoding

minghuaw:
related: #136


minghuaw:
Closing. Will make another PR with a cleaner git history

PR #175, opened 2025-01-12
chore: bump and unify version

minghuaw:
@dcvz Shall we use the workspace version (whatever upstream `mlx` version that `mlx-c` depends on, `"0.21.0"` in this case) for all but `mlx-sys`? And `mlx-sys` will follow the version of `mlx-c` (`"0.1.0"` in this case).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,9 +1,18 @@
 [workspace.package]
-version = "0.14.0"
+# All but mlx-sys should follow the same version. mlx-sys should follow 
+# the version of mlx-c.
+version = "0.21.0"
+edition = "2021"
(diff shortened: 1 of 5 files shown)


minghuaw:
> @minghuaw yeah, I'd agree with that. We'll have some cases where we make Rust side fixes while still depending on the same MLX version but I think we could likely get away with using `-X` to denote that. Something like `0.21.0-3` denoting that this is the 3rd release from our end that uses MLX `0.21.0`

I think anything after the patch version, ie. `0.21.0-xxx` is treated as pre-release and won't be the default when users do `cargo add mlx-rs`

PR #176, opened 2025-01-15
feat: Add conversion between `safetensors::tensor::TensorView` and `Array`

minghuaw:
We should probably document the difference between this conversion and directly loading/saving a `.safetensors` file


minghuaw:
> If anything though we can at least document the try from methods to reflect that

My concern is that, at least for me, I don't often scroll all the way down to the trait impl section to see the documentations. So what I was picturing is a single section of documentation that (briefly) talks about both io and conversion? I'm not sure if this is the best way tho

PR #177, opened 2025-01-15
feat: Allow `#[param(rename = "<rename>"]`

minghuaw:
@dcvz Should we add this though?


minghuaw:
> @minghuaw I honestly wouldn't until there's a bigger demand for it or we find its a crucial and often needed workaround.

👍 

I was feeling the same and just wasn't sure. I'll close this for now.

PR #205, opened 2025-02-24
chore: Update changelog to include changes in #204

minghuaw:


minghuaw:
> > @dcvz A quick question. Should we have separate changelogs for each sub-crates or a single one for the whole project?
> 
> I would do it for the whole project. I think that's easier to surface and that's what we'd have on Github release pages. What do you think?

Agree. Let's do this in the next release.

PR #213, opened 2025-03-27
fix: Fix wrong condition in smooth l1 loss

minghuaw:
Trying to merge this into main because it's a bug fix and not a breaking change. @dcvz WDYT?

See also: https://github.com/ml-explore/mlx-swift/pull/209


minghuaw:
> @minghuaw could we cover this with a test though?

Good point. Let me add one later this week.

minghuaw:
@dcvz I've added a unit test. If you're okay with it, I'll go ahead and merge this and create a patch release

PR #219, opened 2025-04-02
feat: Place state as the last argument in `compile_with_state`

minghuaw:
This PR moves state to the last argument in `compile_with_state`. This is to be consistent with other functions that have a `_with_x` variant where the `x` arg is often placed as the last argument


minghuaw:
Nvm, keeping `state` as the first argument makes `compile_with_state` consistent with the primary usage of this function where the model is the state, and `nn::value_and_grad` requires the model to be the first argument.

I'm closing this then

PR #262, opened 2025-06-06
feat: Add scoped default stream

### LINKED ISSUE #257: Support scoped default stream
See https://github.com/ml-explore/mlx-swift/blob/d3f89b1f520ea9e52fd9de68c1dca708f040406a/Source/MLX/Stream.swift#L92

minghuaw:
Closes #257

### DIFF
--- a/mlx-internal-macros/src/lib.rs
+++ b/mlx-internal-macros/src/lib.rs
@@ -82,13 +82,16 @@ pub fn default_device(attr: TokenStream, item: TokenStream) -> TokenStream {
     // Prepend default stream initialization
     let default_stream_stmt = match input.map(|input| input.device) {
         Some(DeviceType::Cpu) => parse_quote! {
-            let stream = crate::StreamOrDevice::cpu();
+            let stream = crate::task_local_default_stream()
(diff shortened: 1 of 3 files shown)


mlx-internal-macros/src/lib.rs
```diff
@@ -82,13 +82,13 @@ pub fn default_device(attr: TokenStream, item: TokenStream) -> TokenStream {
     // Prepend default stream initialization
     let default_stream_stmt = match input.map(|input| input.device) {
         Some(DeviceType::Cpu) => parse_quote! {
-            let stream = crate::StreamOrDevice::cpu();
+            let stream = crate::Stream::task_local_or_cpu();
```
minghuaw, line 85: @dcvz Not sure if it's a good decision to use `Stream::task_local_or_cpu/gpu/default()` in all generated fn variants? The introduction of `with_new_default_stream` brings scoped task local default stream. `Stream::task_local_or_cpu/gpu/default()` returns the task local default stream if present, otherwise returns the corresponding cpu/gpu/default stream.

PR #267, opened 2025-07-01
Adding mlx_real and mlx_imag functions support.

dshan4585:
This commit improves support on complex arrays by including `mlx_real` & `mlx_imag` functions that is implemented in `mlx-c` library.

```rust
let array = Array::from_complex64(complex::new(0.0, 1.0));

// array.real() now should return array(0, dtype=float32)
// array.imag() now should return array(1, dtype=float32)
```


dshan4585:
I just added unit tests identical to the example. Thanks!

minghuaw (approved):
LGTM! 👍

PR #273, opened 2025-07-15
Prevent premature destructuring for new_mlx_{closure, fallible_closure}

dshan4585:
I am submitting a fix to prevent premature destructuring in the `new_mlx_closure` and `new_mlx_fallible_closure` functions located in `src/util/mod.rs`.

For context, I am planning to add support for custom metal kernel functionality (similar to [python](https://ml-explore.github.io/mlx/build/html/python/_autosummary/mlx.core.fast.metal_kernel.html) and [swfit](https://github.com/ml-explore/mlx-swift/blob/main/Source/MLXFast/MLXFastKernel.swift) versions), which can theoretically be implemented through mlx_sys (the code is dirty and not yet ready to be shared).

The major bug I encountered was related to using the move keyword for my custom closure and passing it to new_mlx_closure-related functions. By acquiring the mlx_closure with `https://github.com/oxideai/mlx-rs/blob/dev/mlx-rs/src/utils/mod.rs`, the implementation uses:
```rust
    let mut closure = Box::from_raw(raw_closure);
```
This closure is almost guaranteed to be dropped at the end of its scope. This behavior makes the generated mlx_closure  unsafe to be reused multiple times, because if the Rust closure holds moved variables, it is deallocated immediately, and my kernel will be lost after single run.

Therefore, I propose adding guards to prevent the premature destruction of the closure. For example:
```rust
extern "C" fn trampoline<'a, F>(
    ret: *mut mlx_vector_array,
    vector_array: mlx_vector_array,
    payload: *mut std::ffi::c_void,
) -> i32
where
{
    unsafe {
        let raw_closure: *mut F = payload as *mut _;
        // Let the box take care of freeing the closure
        let mut closure = Box::from_raw(raw_closure);
        let arrays = match mlx_vector_array_values(vector_array) {
            Ok(arrays) => arrays,
            Err(_) => {
                let _ = Box::into_raw(closure); // prevent premature drop 
                return FAILURE;
            }
        };
        let result = closure(&arrays);
        let _ = Box::into_raw(closure); // prevent premature drop 

        // We should probably keep using new_mlx_vector_array here instead of VectorArray
        // since we probably don't want to drop the arrays in the closure
        *ret = new_mlx_vector_array(result);

        SUCCESS
    }
}
```

Additionally, I suggest adding a manual closure destructor to be called at the end of the mlx_closure's lifetime:
```rust
// extern "C" fn noop_dtor(_data: *mut std::ffi::c_void) {}

extern "C" fn closure_dtor<F>(payload: *mut std::ffi::c_void) {
    if payload.is_null() { return; }
    unsafe {
        drop(Box::from_raw(payload as *mut F));
    }
}
```

After implementing this fix, I was able to write my desired functionality. I believe this change could also relax the strict pure-function requirement for closure-related functions and compilation.

### DIFF
--- a/mlx-rs/src/utils/mod.rs
+++ b/mlx-rs/src/utils/mod.rs
@@ -256,7 +256,7 @@ where
     let payload = raw as *mut std::ffi::c_void;
 
     unsafe {
-        mlx_sys::mlx_closure_new_func_payload(Some(trampoline::<F>), payload, Some(noop_dtor))
+        mlx_sys::mlx_closure_new_func_payload(Some(trampoline::<F>), payload, Some(closure_dtor::<F>))
     }
 }
 
@@ -272,7 +272,7 @@ where
         mlx_sys::mlx_closure_new_func_payload(
             Some(trampoline_fallible::<F>),
             payload,
-            Some(noop_dtor),
+            Some(closure_dtor::<F>),
         )
     }
 }
@@ -315,10 +315,13 @@ where
         let arrays = match mlx_vector_array_values(vector_array) {
             Ok(arrays) => arrays,
             Err(_) => {
+                let _ = Box::into_raw(closure); // prevent premature drop 
                 return FAILURE;
             }
         };
         let result = closure(&arrays);
+        let _ = Box::into_raw(closure); // prevent premature drop 
+
         // We should probably keep using new_mlx_vector_array here instead of VectorArray
         // since we probably don't want to drop the arrays in the closure
         *ret = new_mlx_vector_array(result);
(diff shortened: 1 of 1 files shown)


minghuaw:
Got sick. Will review this later

minghuaw:
> For context, I am planning to add support for custom metal kernel functionality

Really looking forward to this 👍

minghuaw:
> I think this PR might be missing from the current dev branch. @minghuaw , were any issues found with this PR after it was merged?

Good catch. I honestly don't remember. Would you mind submitting these changes in another PR.

dshan4585:
> > I think this PR might be missing from the current dev branch. @minghuaw , were any issues found with this PR after it was merged?
> 
> Good catch. I honestly don't remember. Would you mind submitting these changes in another PR.

Thanks for the reply. I'll make another PR.

PR #288, opened 2025-10-19
Add missing #[param] annotations to QuantizedEmbedding

tokav:
Because QuantizedEmbedding struct is missing `#[param]` annotation on its `scales`, `biases` and `inner` it causes `Embedding` quantization to miss these parameters entirely when loading quantized model weights.


tokav:
In my PoC project I manually patched it + I added bunch of "mlx_lm" code (along with more models support, quantization, context size etc...) that you can use however you want: [mlx-lm-poc](https://github.com/tokav/mlx-lm-poc).

minghuaw:
Good catch. Thanks!

PR #296, opened 2025-12-02
Fix not one but two leaks related to gradients

### LINKED ISSUE #290: Resource leak when evaluating gradient
To reproduce, run the following program:
```toml
[dependencies]
mlx-rs = { version = "0.25.2", features = [ "metal" ], default-features = false }
```
```rust
use mlx_rs::Array;
use mlx_rs::error::Exception;

fn main() {
    fn f(x: &Array) -> Result<Array, Exception> {
        x.multiply(x)
    }
    let mut grad_f = mlx_rs::transforms::grad(f);

    let x: Array = 1.0.into();
    for i in 0.. {
        dbg!(i);
        let gf = grad_f(&x).unwrap();
        gf.eval().unwrap();
    }
}
```

On each iteration it creates one more Mach port (as can be seen in macOS Activity Monitor, in the "Memory" tab).
And eventually it fails with
```
...
[src/main.rs:12:9] i = 65516
[src/main.rs:12:9] i = 65517

thread 'main' (2298976) panicked at src/main.rs:14:19:
called `Result::unwrap()` on an `Err` value: Exception { what: "[Event::stream] Cannot access stream on invalid event. at /Users/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mlx-sys-0.2.0/src/mlx-c/mlx/c/array.cpp:324", location: Location { file: "/Users/user/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mlx-rs-0.25.2/src/array/mod.rs", line: 302, column: 9 } }
```

Vlad-Shcherbina:
(fixes #290)

Please see individual commit messages for more details.

Missing `mlx_closure_value_and_grad_free()` causes relatively minor memory leak.
Missing `mlx_vector_array_free()` leaks both memory and Mach ports.

With both these fixes, the loop from the issue description can run indefinitely.

### DIFF
--- a/mlx-rs/src/utils/guard.rs
+++ b/mlx-rs/src/utils/guard.rs
@@ -144,7 +144,9 @@ impl Guard<Vec<Array>> for MaybeUninitVectorArray {
         self.init_success = success;
     }
 
-    fn try_into_guarded(self) -> Result<Vec<Array>, Exception> {
+    fn try_into_guarded(mut self) -> Result<Vec<Array>, Exception> {
+        debug_assert!(self.init_success);
+        self.init_success = false; // mlx_vector_array still needs to be freed after we extracted its elements
         unsafe {
             let size = mlx_sys::mlx_vector_array_size(self.ptr);
             (0..size)


minghuaw:
Nice! LGTM! Thanks a lot for fixing this

PR #302, opened 2025-12-29
Add Sync trait to Array with safety documentation

pjm073:
MLX uses atomic reference counting for arrays, making them thread-safe for concurrent access. This adds the Sync trait with comprehensive safety documentation and tests.

Changes:
- Add unsafe impl Sync for Array with detailed safety comments
- Add comprehensive thread safety tests:
  - test_array_send: Verify arrays can be sent between threads
  - test_array_sync: Verify arrays can be shared via Arc
  - test_concurrent_operations: Test concurrent array operations
  - test_no_memory_leak_gradient: Regression test for memory leaks
- Remove outdated TODO comment in Drop implementation
- Document MLX's atomic reference counting guarantees

The Sync trait is safe to implement because:
1. MLX uses atomic reference counting internally
2. The underlying C++ implementation is thread-safe
3. Multiple threads can hold references and call methods concurrently

These tests verify that arrays work correctly when:
- Moved between threads (Send)
- Shared between threads via Arc (Sync)
- Used in concurrent operations from multiple threads

### DIFF
--- a/mlx-rs/src/array/mod.rs
+++ b/mlx-rs/src/array/mod.rs
@@ -64,14 +64,20 @@ impl std::fmt::Display for Array {
 
 impl Drop for Array {
     fn drop(&mut self) {
-        // TODO: check memory leak with some tool?
-
         // Decrease the reference count
         unsafe { mlx_sys::mlx_array_free(self.as_ptr()) };
     }
 }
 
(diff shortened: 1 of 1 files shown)


pjm073:
Closing - created this by mistake. Will reopen when ready to contribute upstream.

PR #306, opened 2026-01-12
feat: add TrainableState wrapper for efficient JIT compilation

nicholasjpaterno:
## Summary

When using `compile_with_state` with LoRA or other parameter-efficient fine-tuning methods, the standard `Updatable` implementation returns ALL model parameters. For a model with 10M parameters but only 500K trainable, this causes MLX's compiler to prune 9.5M unchanged arrays from output, breaking state tracking.

`TrainableState` wraps a model and optimizer, implementing `Updatable` to return only trainable (non-frozen) parameters plus optimizer state. This dramatically reduces state count, enabling successful JIT compilation.

## Changes

- `TrainableState<M, O>` wrapper type
- `CompileTrainingExt` trait for ergonomic wrapping
- Consistent state ordering via cached trainable keys

## Example Usage

```rust
use mlx_rs::transforms::compile::{compile_with_state, CompileTrainingExt};

// Wrap model and optimizer
let trainable = (model, optimizer).trainable_state();

// Compile - only trainable params in state
let compiled_step = compile_with_state(training_step, trainable)?;
```

## Test Plan

- Tested with LoRA fine-tuning on 0.6B+ parameter models
- Verified state count matches trainable parameter count
- Confirmed JIT compilation succeeds where it previously failed

### DIFF
--- a/mlx-rs/src/transforms/compile/mod.rs
+++ b/mlx-rs/src/transforms/compile/mod.rs
@@ -151,9 +151,11 @@ use crate::Array;
 #[allow(clippy::module_inception)]
 mod compile;
 mod compile_with_state;
+mod trainable_state;
 
 pub use compile::*;
 pub use compile_with_state::*;
+pub use trainable_state::*;
 
 /// Globally enable the compilation of functions.
 ///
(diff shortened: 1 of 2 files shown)


minghuaw:
This might be a bug with the Updatable impl. IIRC, `Updatable` is supposed to achieve the same goal. I'll need to check that

nicholasjpaterno:
@dcvz yes I think you're right. Happy to give #314 a test run

PR #310, opened 2026-01-25
Add contiguity check to try_as_slice() and contiguous() function

ymote:
## Summary

- Add `is_contiguous()` method to `Array` to check if the array has contiguous memory layout
- Add `NotContiguous` error variant to `AsSliceError`
- Update `try_as_slice()` to check contiguity before returning slice (BREAKING CHANGE)
- Add `contiguous()` function to `ops` module to make non-contiguous arrays contiguous

## Motivation

Previously, `as_slice()` would return raw memory ignoring strides, causing silent data corruption when the array was not contiguous. This was a significant source of bugs when working with indexed or transposed arrays.

For example:
```rust
let arr = Array::from_slice(&[1i32, 2, 3, 4, 5, 6], &[2, 3]);
let transposed = ops::transpose(&arr)?;
let slice = transposed.as_slice::<i32>();  // WRONG DATA!
```

The slice would contain elements in the wrong order because `transpose()` creates a strided view, but `as_slice()` ignored the strides.

### `Array::is_contiguous()`
Checks if the array is contiguous (row-major/C-style) by verifying strides match expected values.

### `AsSliceError::NotContiguous`
New error variant returned when calling `try_as_slice()` on a non-contiguous array.

### `try_as_slice()` update
Now checks contiguity after evaluation and returns `NotContiguous` error if the array is not contiguous.

### `ops::contiguous()`
New function that returns a contiguous copy of the array if needed.

## Migration Guide

If your code used `as_slice()` on indexed/transposed arrays, it was likely producing incorrect results. To fix:

```rust
// Option 1: Use contiguous()
let c = ops::contiguous(&arr)?;
let slice = c.as_slice::<f32>();

// Option 2: Reshape to force contiguity
let n = arr.size() as i32;
let c = arr.reshape(&[n])?.reshape(arr.shape())?;
let slice = c.as_slice::<f32>();
```

## Test Plan

- [x] Unit tests for `is_contiguous()` with various array shapes
- [x] Unit tests for `try_as_slice()` returning `NotContiguous` error
- [x] Unit tests for `contiguous()` function
- [x] Verified fix in real-world use case (image generation pipeline that was producing tiled images due to this bug)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dcvz:
Thanks @ymote, this is now present in `main` and will go out in the next few days

PR #305, opened 2026-01-12
fix: prevent integer underflow in compile_with_state for large models

nicholasjpaterno:
## Summary

When using `compile_with_state` with very large models (10M+ parameters), the MLX compiler may prune state arrays that appear unchanged during computation. This caused an integer underflow when calculating the `suffix_start` index, leading to a panic with an out-of-range slice index.

## Changes

- Changed the unsafe subtraction to use `checked_sub()` 
- Returns a descriptive error when the state count doesn't match expectations
- The error message guides users to use non-compiled training for large models or ensure all trainable parameters are being updated

## Test Plan

- Tested with large LoRA fine-tuning scenarios where state pruning occurred
- Verified the error message is helpful for debugging

PR #314, opened 2026-01-28
fix(compile): prevent crash with frozen parameters in compile_with_state

dcvz:
Fix root cause of #305 and #306 - with large models, MLX may prune unchanged arrays from compiled output. The old code used subtraction to find the split point, causing underflow when output had fewer arrays than expected. Now we capture the function output count during tracing and use that as the split point instead.


nicholasjpaterno:
@dcvz I can confirm this solves the issue and obviates the other PRs 🔥

PR #319, opened 2026-02-16
Support Llama 3.2 1B instruct

LYZJU2019:
This PR attempts to support `meta-llama/Llama-3.2-1B-Instruct`

Reference implementation: https://github.com/ml-explore/mlx-examples/blob/main/llms/llama/llama.py

To test, run `hf download meta-llama/Llama-3.2-1B-Instruct` and the model folder should appear in `~/.cache/huggingface/hub/models--meta-llama--Llama-3.2-1B-Instruct/snapshots/<commit_hash>`

### DIFF
--- a/mlx-lm/src/utils/rope.rs
+++ b/mlx-lm/src/utils/rope.rs
@@ -173,31 +173,59 @@ pub enum RopeVariant {
 // TODO: support derive ModuleParameters for enum
 impl mlx_rs::module::ModuleParameters for RopeVariant {
     fn num_parameters(&self) -> usize {
-        0
+        match self {
+            RopeVariant::Default(rope) => rope.num_parameters(),
+            RopeVariant::Llama3(rope) => rope.num_parameters(),
+        }
     }
 
-    fn freeze_parameters(&mut self, _recursive: bool) {}
+    fn freeze_parameters(&mut self, _recursive: bool) {
+        match self {
+            RopeVariant::Default(rope) => rope.freeze_parameters(_recursive),
+            RopeVariant::Llama3(rope) => rope.freeze_parameters(_recursive),
+        }
+    }
 
-    fn unfreeze_parameters(&mut self, _recursive: bool) {}
+    fn unfreeze_parameters(&mut self, _recursive: bool) {
+        match self {
+            RopeVariant::Default(rope) => rope.unfreeze_parameters(_recursive),
+            RopeVariant::Llama3(rope) => rope.unfreeze_parameters(_recursive),
+        }
+    }
 
     fn parameters(&self) -> mlx_rs::module::ModuleParamRef<'_> {
-        mlx_rs::nested::NestedHashMap::new()
+        match self {
+            RopeVariant::Default(rope) => rope.parameters(),
+            RopeVariant::Llama3(rope) => rope.parameters(),
+        }
     }
 
     fn parameters_mut(&mut self) -> mlx_rs::module::ModuleParamMut<'_> {
-        mlx_rs::nested::NestedHashMap::new()
+        match self {
+            RopeVariant::Default(rope) => rope.parameters_mut(),
+            RopeVariant::Llama3(rope) => rope.parameters_mut(),
+        }
     }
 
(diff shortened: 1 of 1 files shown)


mlx-lm/src/utils/rope.rs
```diff
@@ -25,13 +26,224 @@ impl FloatOrString {
...
+        Ok(Self {
+            dimensions: dims,
+            traditional,
+            scale: 1.0,
+            freqs: freqs_array,
+        })
+    }
+}
+
+impl mlx_rs::module::ModuleParameters for Llama3Rope {
```
minghuaw: This can be derived

mlx-lm/src/utils/rope.rs
```diff
@@ -25,13 +26,224 @@ impl FloatOrString {
...
+                let smooth_factor = (old_context_len / wavelen - low_freq_factor)
+                    / (high_freq_factor - low_freq_factor);
+                let smooth_freq = freq / ((1.0 - smooth_factor) / factor + smooth_factor);
+                scaled_freqs.push(smooth_freq);
+            } else {
+                scaled_freqs.push(freq);
+            }
+        }
+
+        let freqs_array = Array::from_slice(&scaled_freqs, &[half_dims]);
```
minghuaw: Instead of converting to mlx array at the end, it might be better to start using mlx ops from the beginning

mlx-lm/src/utils/rope.rs
```diff
@@ -25,13 +26,224 @@ impl FloatOrString {
...
+/// Llama3-style RoPE with frequency scaling.
+///
+/// Applies piecewise frequency scaling based on wavelength cutoffs derived from
+/// `low_freq_factor`, `high_freq_factor`, `factor`, and `original_max_position_embeddings`.
+#[derive(Debug, Clone)]
+pub struct Llama3Rope {
+    pub dimensions: i32,
+    pub traditional: bool,
+    pub scale: f32,
+    pub freqs: Array,
```
minghuaw, line 70: Need to double check whether freqs here is a module parameter (not important in inference, but still worth checking)

mlx-lm/src/utils/rope.rs
```diff
@@ -25,13 +26,224 @@ impl FloatOrString {
...
+
+/// Enum wrapping different RoPE variants so that `initialize_rope` can return
+/// either a standard RoPE or a Llama3 RoPE.
+#[derive(Debug, Clone)]
+pub enum RopeVariant {
+    Default(nn::Rope),
+    Llama3(Llama3Rope),
+}
+
+impl mlx_rs::module::ModuleParameters for RopeVariant {
```
minghuaw, line 174: todo: support derive ModuleParameters for enum

mlx-lm/src/utils/rope.rs
```diff
@@ -25,13 +26,224 @@ impl FloatOrString {
     }
 }
 
+fn get_float_from_config(
```
minghuaw: str variant aren't just float point numbers in string format, there could be values like "default" and "linear" for example.

mlx-lm/src/utils/rope.rs
minghuaw: `freqs` should not be a module param

mlx-lm/src/utils/rope.rs
```diff
@@ -25,13 +34,206 @@ impl FloatOrString {
...
+#[derive(Debug, Clone)]
+pub enum RopeVariant {
+    Default(nn::Rope),
+    Llama3(Llama3Rope),
+}
+
+// TODO: support derive ModuleParameters for enum
+impl mlx_rs::module::ModuleParameters for RopeVariant {
+    fn num_parameters(&self) -> usize {
+        0
```
minghuaw: It's better to just call the implementation of the variant like

```rust
match self {
    RopeVariant::Default(rope) => rope.num_parameters(),
    RopeVariant::Llama3(rope) => rope.num_parameters(),
}
```

minghuaw: Same thing for other functions in this trait

minghuaw (commented):
Just a minor comment. Then it'll be ready to merge

PR #328, opened 2026-03-19
Fix metallib path for `cargo install` — use stable `~/.mlx/` directory

### LINKED ISSUE #327: `cargo install` breaks: metallib path baked in at compile time points to temp directory
## Summary

When a binary using mlx-rs is installed via `cargo install <crate>`, it fails at runtime with:

```
MLX error: Failed to load the default metallib. library not found library not found library not found
```

The issue is that `mlx-sys`'s CMake build defines `METAL_PATH` as an absolute path to the metallib inside Cargo's temp build directory:

```
-DMETAL_PATH="/var/folders/.../T/cargo-installXXXX/release/build/mlx-sys-.../mlx.metallib"
```

This path is compiled into `device.cpp.o`. When `cargo install` finishes, only the binary is copied to `~/.cargo/bin/` — the temp build dir (and the metallib in it) is deleted.

## MLX's metallib search order

From `device.cpp`, MLX searches in order:
1. Co-located with the binary (`get_binary_directory()` / `mlx.metallib`)
2. Bundle resources
3. SwiftPM bundle
4. Compile-time `METAL_PATH` (the broken fallback)

## Workaround

Copying the metallib next to the installed binary works because of search path #1:

```bash
cp target/release/build/mlx-sys-*/out/build/_deps/mlx-build/mlx/backend/metal/kernels/mlx.metallib ~/.cargo/bin/
```

Building from source (where `target/` persists) also works fine because `METAL_PATH` still points to a valid location.

## Potential fix

The `build.rs` in mlx-sys could copy `mlx.metallib` to a location that `cargo install` preserves. However, `cargo install` only copies the final binary — it doesn't preserve any other build artifacts.

Some ideas:
- Use `include_bytes!` to embed the metallib in the binary and write it to disk at runtime (it's ~84MB though, so this bloats the binary significantly)
- Set `METAL_PATH` to a well-known stable location like `~/.local/share/mlx/mlx.metallib` and copy it there during build
- Document that downstream crates using mlx-rs should be installed from source

## Environment

- macOS 26.3 (arm64, Apple Silicon)
- Rust 1.93.0
- mlx-rs 0.25.3 / mlx-sys 0.2.0

## Reproduction

```bash
cargo install voice  # any crate that uses mlx-rs
voice "hello"        # fails with metallib error
```

This affects the [voice](https://crates.io/crates/voice) TTS crate (and likely any binary crate using mlx-rs).

rgbkrk:
## Problem

When a binary using mlx-rs is installed via `cargo install`, it fails at runtime:

```
MLX error: Failed to load the default metallib. library not found library not found library not found
```

`cargo install` builds in a temporary directory, then copies only the binary to `~/.cargo/bin/`. The CMake build bakes `METAL_PATH` as an absolute path to the metallib inside this temp dir. After install, the temp dir is deleted and the metallib is unreachable.

## Fix

Override `MLX_METAL_PATH` in the CMake config to point to a stable versioned directory: `~/.mlx/lib/v{mlx-sys-version}/`. The directory is created before CMake runs so CMake outputs the compiled metallib there directly. The compiled-in `METAL_PATH` now survives temp dir cleanup.

MLX's runtime metallib search order:
1. Co-located with binary → still works for local builds
2. Bundle resources → still works
3. SwiftPM bundle → still works
4. **`METAL_PATH` compiled-in fallback → now points to `~/.mlx/lib/v0.2.0/mlx.metallib`** ✅

## Testing

Tested with the [voice](https://github.com/rgbkrk/voicers) TTS crate:

```bash
# Patched build — metallib goes to stable location
cargo build --release -p voice
strings target/release/voice | grep metallib
# Shows: /Users/.../.mlx/lib/v0.2.0/mlx.metallib

ls ~/.mlx/lib/v0.2.0/mlx.metallib
# -rw-r--r--  84M  mlx.metallib

# Install and run — works even though build dir is gone
cargo install --path crates/voice-cli --force
voice -o /tmp/test.wav "hello"  # Success!
```

Fixes #327

Note: this PR is based on the v0.25.3 tag. Happy to rebase onto main if preferred — the `build.rs` on main has additional changes (clang_rt workaround) but the fix is the same.

### DIFF
--- a/mlx-sys/build.rs
+++ b/mlx-sys/build.rs
@@ -1,7 +1,7 @@
 extern crate cmake;
 
 use cmake::Config;
-use std::{env, path::PathBuf, process::Command};
+use std::{env, fs, path::PathBuf, process::Command};
 
 /// Find the clang runtime library path dynamically using xcrun
 fn find_clang_rt_path() -> Option<String> {
@@ -44,6 +44,27 @@ fn find_clang_rt_path() -> Option<String> {
     None
 }
 
+/// Determine a stable directory for the MLX metallib that survives `cargo install`
+/// temp dir cleanup. Uses `~/.mlx/lib/v{version}/`.
+///
+/// When `cargo install` builds a crate, it uses a temporary directory that is
+/// deleted after the binary is copied. The CMake build bakes the metallib path
+/// into the binary via `-DMETAL_PATH=...`. If that path points to the temp dir,
+/// the binary fails at runtime with "Failed to load the default metallib".
+///
+/// By setting MLX_METAL_PATH to a stable home-directory location (and creating
+/// it before CMake runs so CMake can output the metallib there directly), the
+/// compiled-in METAL_PATH remains valid after the temp dir is cleaned up.
+#[cfg(feature = "metal")]
+fn stable_metallib_dir() -> PathBuf {
+    let version = env!("CARGO_PKG_VERSION");
+    let home = env::var("HOME").unwrap_or_else(|_| "/tmp".to_string());
+    PathBuf::from(home)
+        .join(".mlx")
+        .join("lib")
+        .join(format!("v{}", version))
+}
+
 fn build_and_link_mlx_c() {
     let mut config = Config::new("src/mlx-c");
     config.very_verbose(true);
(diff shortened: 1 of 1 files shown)


minghuaw:
Would you please rebase onto the main branch? I'll try to take a look this weekend.

rgbkrk:
done

dcvz:
Thank you! Fixed in #365

PR #346, opened 2026-04-29
feat: add memory::set_memory_limit and memory::get_memory_limit

### LINKED ISSUE #225: feat: Set and get GPU memory limit
See also: https://github.com/ml-explore/mlx-swift/pull/223

dschulmeist:
Closes #225.

## Changes

Adds a new `memory` module under `mlx-rs/src/memory.rs` exposing:

- `memory::get_memory_limit() -> Result<usize>`
- `memory::set_memory_limit(limit: usize) -> Result<usize>` — returns the previous limit so callers can save and restore it.

The module is wired in via `pub mod memory;` in `lib.rs`. The error-handling pattern mirrors `utils/io.rs` (`INIT_ERR_HANDLER` + `get_and_clear_last_mlx_error`).

This gives mlx-rs parity with mlx-swift's `Memory.memoryLimit` getter/setter (referenced by [mlx-swift#223](https://github.com/ml-explore/mlx-swift/pull/223)).

## A note on the issue body

The issue references `mlx_metal_set_memory_limit`, but that symbol no longer exists upstream — the memory limit functions were moved out of the `metal` namespace and the current mlx-c API is `mlx_set_memory_limit` / `mlx_get_memory_limit`. This PR binds those.

## Scoped

Just the memory limit, matching the issue title. The other functions in `mlx/c/memory.h` (`mlx_get_active_memory`, `mlx_get_cache_memory`, `mlx_get_peak_memory`, `mlx_set_cache_limit`, `mlx_clear_cache`, etc.) are out of scope here and a good fit for a follow-up.

## Testing

`cargo test -p mlx-rs --lib memory::` — one combined test asserting:
- Initial `get` returns the upstream default (1.5x recommended max working set on Apple Silicon).
- `set(a)` then `get()` returns `a`.
- Consecutive `set(b)` returns `a` (the previous limit contract).
- Final `get()` reflects `b`, then restore.

The two contracts (round-trip and previous-value) live in one test on purpose: memory limit is global state on the active device, so a parallel test runner could otherwise interleave writes between two separate tests.

### DIFF
--- a/mlx-rs/src/lib.rs
+++ b/mlx-rs/src/lib.rs
@@ -292,6 +292,7 @@ pub mod fast;
 pub mod fft;
 pub mod linalg;
 pub mod losses;
+pub mod memory;
 pub mod module;
 pub mod nested;
 pub mod nn;
--- a/mlx-rs/src/memory.rs
+++ b/mlx-rs/src/memory.rs
@@ -1,0 +1,115 @@
+//! GPU memory management.
+//!
+//! These functions wrap the global memory-management APIs from `mlx-c`. They
+//! affect the active device, which on Apple Silicon is the unified-memory
+//! Metal allocator.
+//!
+//! # Example
+//!
+//! ```rust,no_run
+//! use mlx_rs::memory;
+//!
+//! // Cap MLX's GPU allocator at 8 GiB. Returns the previous limit.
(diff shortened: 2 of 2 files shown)


dcvz:
Thank you! Fixed in #367 — went with the full memory family while at it.

PR #350, opened 2026-05-14
feat(compile): add boxed slice helpers

rabbitson87:
## Summary

Adds boxed compiled-function helpers for slice-based compile users:

- `compile_boxed_slice` for `&[Array]` inputs
- `compile_boxed_slice_refs` for `&[&Array]` inputs
- public type aliases for both boxed closure shapes

This makes it possible to store compiled slice functions behind trait objects, for example in caches or runtime-selected containers, while keeping the compiled state alive until the boxed closure is dropped.

## Validation

- `cargo fmt -- --check`
- `cargo clippy -p mlx-rs --lib -- -D warnings`
- `cargo test -p mlx-rs --lib test_type_id_to_usize -- --test-threads=1`
- `git diff --check`

I also added focused tests for both new helpers. In my local worktree, runtime tests that enter the compile path abort with `Rust cannot catch foreign exceptions`; the existing `test_compile_with_error` aborts the same way, so I could verify compilation/lints locally but not MLX runtime execution.

### DIFF
--- a/mlx-rs/src/transforms/compile/compile.rs
+++ b/mlx-rs/src/transforms/compile/compile.rs
@@ -8,6 +8,13 @@ use crate::{error::Exception, Array};
 
 use super::{type_id_to_usize, Closure, Compiled, CompiledState, Guarded, VectorArray};
 
+/// Type-erased compiled function over an owned array slice.
+pub type BoxedCompiledSlice = Box<dyn FnMut(&[Array]) -> Result<Vec<Array>, Exception> + Send>;
+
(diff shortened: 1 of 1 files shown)


dcvz:
Sorry this sat so long @rabbitson87. The compile internals got reworked quite a bit in the meantime and `Compiled` is intentionally `!Send` now, so this doesn't apply cleanly anymore. Boxing what `compile()` returns should cover the type-erasure case. If that's not enough for what you were doing, open an issue and we'll figure out a helper.

PR #353, opened 2026-05-20
docs(readme): bump MSRV claim to 1.85.0

fiorelorenzo:
Workspace `Cargo.toml` has `rust-version = "1.85.0"` (`Cargo.toml:16`) and CI runs `cargo 1.85` (`.github/workflows/validate.yml`). The README's MSRV section still claimed 1.83.


dcvz:
Thank you! Fixed in #366 (the MSRV had drifted further — it is 1.88.0 now, matching CI).

PR #357, opened 2026-06-18
fix(nn): apply RoPE on the input shape, not a reshaped 3D tensor

sergey-scherbina:
## Problem

`RotaryPositionalEncoding::forward` reshapes `[B, n_heads, L, head_dim]` to
`[-1, L, head_dim]` before calling `mx.fast.rope`, then reshapes back. For
single-position input (`L == 1`, i.e. every autoregressive decode step) the
collapsed `[B*n_heads, 1, head_dim]` shape hits a `fast::rope` bug that rotates
only the first row along the leading axis — every head past the first is left
**un-rotated**. The result is corrupt decode queries and garbage generation.

Prefill (`L > 1`) is unaffected, so this only surfaces during incremental
decoding, which makes it easy to miss.

Python `mlx_lm` applies `mx.fast.rope` to the 4D tensor directly and is correct.

## Fix

Drop the reshape and apply RoPE on the input shape directly.

## Test

Adds `test_rope_rotates_all_heads_when_single_position`: feeds four identical
heads at a non-zero offset and asserts every head comes out identical to a
single-head reference (RoPE depends only on position, so identical heads must
rotate identically) and that the rotation is non-trivial. Fails before this
change (heads 1–3 stay un-rotated), passes after.

### DIFF
--- a/mlx-rs/src/nn/positional_encoding.rs
+++ b/mlx-rs/src/nn/positional_encoding.rs
@@ -119,9 +119,13 @@ where
 
     fn forward(&mut self, input: Input) -> Result<Self::Output, Self::Error> {
         let RopeInput { x, offset } = input.into();
-        let shape = x.shape();
-        let x = x.reshape(&[-1, x.dim(-2), x.dim(-1)])?;
-        let x = crate::fast::rope(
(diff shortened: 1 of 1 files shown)


dcvz:
Thank you! Included in #365

PR #360, opened 2026-08-04
build: update pmetal bindings to MLX 0.32

ilblackdragon:
## Summary
- advance the published pmetal binding lineage from MLX 0.30.6 to official MLX 0.32.0
- preserve mlx-c 0.5 ABI while supplying MLX 0.32 FFT-normalization and quantization global-scale defaults
- make implicit Rust RNG state thread-local to match MLX 0.32 thread-affine streams
- add a cross-thread regression test for leaked RNG stream handles

The branch is based on the commits used to publish `pmetal-mlx-rs 0.25.8` / `pmetal-mlx-sys 0.2.4`, which are newer than the current default branch.

## Validation
- `cargo test -p pmetal-mlx-rs random::tests:: -- --test-threads=1` (32 passed)
- downstream HoloNear MLX backend suite (550 passed, 23 ignored)
- confirmed staged native source fetches exact tag `v0.32.0`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,10 +32,10 @@ resolver = "2"
 
 [workspace.dependencies]
 # workspace local dependencies
-mlx-sys = { version = "=0.2.3", path = "mlx-sys", package = "pmetal-mlx-sys" }
+mlx-sys = { version = "=0.2.4", path = "mlx-sys", package = "pmetal-mlx-sys" }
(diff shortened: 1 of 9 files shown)


dcvz:
Thanks @ilblackdragon, and sorry for the silence on this one. We ended up going straight to 0.32.2 on main (#362) so I'll close this, but your RNG thread-exit regression idea made it into #366. The memory module will also be handy once we get to the memory limit APIs.

